Issue #23, opened 2014-11-23
No specs in time module


nmnowak:
The date module seems to be really well speced out but I don't see any specs in the time module. Is there any reason for this or is it just something that is waiting to be done, maybe by me?

X4lldux:
Plus there are some missing type information in some modules.

bitwalker:
I've manually merged @gabrielelana's PR adding some types/specs, so I'll close this for now, as fleshing out documentation is part of #9

PR #30, opened 2014-12-31
Add docs and specs in Time module

gabrielelana:
Addressing issue #23

Focus will be to keep dialyzer output clean for the Time module, aka the output of the following command needs to be empty
```
dialyzer --no_check_plt --plt /home/coder/.dialyxir_core_17_1.0.2.plt -Wunmatched_returns -Werror_handling -Wrace_conditions -Wunderspecs /home/coder/code/timex/_build/dev/lib/timex/ebin | ack -i ‘^time.ex’
```

I will also fix bugs (if any) found by dialyzer

### DIFF
--- a/lib/time/time.ex
+++ b/lib/time/time.ex
@@ -2,29 +2,61 @@ defmodule Timex.Time do
 
   @million 1_000_000
 
+  @type units :: :usecs | :msecs | :secs | :mins | :hours | :days | :weeks | :hms
+  @type quantity :: float
+
+
   Enum.each [usecs: @million, msecs: 1_000], fn {type, coef} ->
+    @spec to_usecs(quantity, unquote(type)) :: quantity
     def to_usecs(value, unquote(type)), do: value * @million / unquote(coef)
+
+    @spec to_msecs(quantity, unquote(type)) :: quantity
     def to_msecs(value, unquote(type)), do: value * 1000 / unquote(coef)
+
+    @spec to_secs(quantity, unquote(type)) :: quantity
     def to_secs(value, unquote(type)),  do: value / unquote(coef)
+
+    @spec to_mins(quantity, unquote(type)) :: quantity
     def to_mins(value, unquote(type)),  do: value / unquote(coef) / 60
+
+    @spec to_hours(quantity, unquote(type)) :: quantity
     def to_hours(value, unquote(type)), do: value / unquote(coef) / 3600
+
+    @spec to_days(quantity, unquote(type)) :: quantity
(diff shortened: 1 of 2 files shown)


bitwalker:
@gabrielelana Beautiful! Do you want me to merge this as is, or wait for further improvements?

gabrielelana:
Sorry, I've been a little bit busy, I'll continue to work on this :smile:

bitwalker:
No worries, I know how it goes :). Whenever you want me to take a look, just ping me here!

gabrielelana:
@bitwalker I discovered that it’s not so easy to add specs/docs to some code generated by a macro :smile: 

docs_rendering

As you can see the specs doesn’t look good, should be something like
```elixir
to_msecs(quantity, :usecs | :msecs | :secs | :mins | :hours | :days | :weeks) :: quantity
to_msecs({quantity, quantity, quantity}, :hms) :: quantity
```

I don’t see how to do that with the current code.

Moreover the doc AFAIK it’s impossibile to write because, unlike the specs, it’s impossibile to write cumulative docs for the same function (aka same name, same arity), in the picture you see “Converts weeks to milliseconds” because the conversion from weeks is the last to be generated

I deleted the doc after the rendering because was useless

What do you think we should do? Should we organize the code in a different way? Maybe generating all the clauses of a function in sequence?

```
for each function f
  @doc for f
  @spec for f
  for each unit u
    define f for u
  end
end
```

bitwalker:
Oh I didn't even think about that. You can define a function head (no body) for the name/arity pair, prior to any of the implementations, and that's where you decorate the `@doc` and `@spec` attributes. It looks like this:

```elixir
@doc """
Documentation for foo
"""
@spec foo(:baz | :qux) :: {:ok, binary} | {:error, binary}
def foo(type)

def foo(:baz), do: {:ok, "baz"}
def foo(:qux), do: {:ok, "qux"}
def foo (_),   do: {:error, "Invalid type"}
```

Stupid example, but hopefully that gives you the gist. So applying it to the issue you mentioned, *should* be as simple as defining the function heads prior to the implementation definitions. I hope anyways. Let me know if that ends up not being the case and we can think about this a bit more.

bitwalker:
Refactoring as you mentioned is also another approach we could take, or perhaps it will require a combination of the two solutions. I'll let you decide what you think is cleanest, and we can go from there.

bitwalker:
Oh and damn, I almost forgot, but you might want to rebase on master, as I just merged a PR which adds support for `mix docs`, which you may find useful when working on the documentation.

bitwalker:
I've manually merged this into master. Thanks!

PR #32, opened 2015-01-09
Date to secs bug 24

bbense:
This patch fixes the Date.to_secs bug and also adds some additional tests that came up during fixing the problem. There were several places in date.ex that had code that broke when this bug was fixed. This current code passes all the existing tests. 

I also added some documentation and the deps needed for mix docs to produce documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
-docs/
+doc/
+doc
 deps/
 ebin/
 _build/
 .exenv-version
+mix.lock
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -881,15 +881,9 @@ defmodule Timex.Date do
   def compare(_, :distant_future), do: -1
   def compare(date, date),         do: 0
   def compare(a, b),               do: compare(a, b, :secs)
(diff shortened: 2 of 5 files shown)


bitwalker:
This is all great stuff! Thanks for all the effort!

Since it's going to be rather hard to split these commits up into different PRs, I'll go ahead and merge as-is. But if possible in the future, try and create separate PRs for different issues/features - that way if you fix a bug, but there is a feature addition which I may want to have some discussion on, I can merge the bugfix, but keep the feature addition PR open for a bit. No worries on this one, just something to keep in mind.

I had originally intended on implementing Read The Docs for Timex, but I think what you've done here is a better choice for now, and gets documentation into people's hands today instead of some nebulous future date, so I thank you for taking the lead on that! At some point in the future I may implement an additional task for publishing to RTD, but that's something for another day.

bbense:
Yeah, I got a bit carried away. But I thought the changes minimal enough
that it would not be to problematic
to include them and I needed the docs to help debug the problem.

- Booker C. Bense

bitwalker:
Yeah it's not a big deal, and in the case wasn't an issue anyways. If you look through the commit history, there is no shortage of times where I've broken my own rules, so I can't hold it against anyone else if they do the same ;).

PR #34, opened 2015-01-12
Fix incorrect unit names in the docs for Timex.Time functions

alco:
This patch fixes outdated docs.

Nowadays, I would actually use full names for the units. Those would be simpler and more intuitive for the users: `:secs` -> `:seconds`, `:mins` -> `:minutes`, `:usecs` -> `:microseconds`, etc. Users are likely to utilize just one or two units, so they can wrap those long names in a function for convenience, like so:

```elixir
defmodule MyApp.Utils do
  @doc "Returns the difference between two timestamps in microseconds"
  @spec time_diff(Timex.Date.timestamp, Timex.Date.timestamp) :: Timex.Date.microseconds
  def time_diff(timestamp1, timestamp2) do
    Timex.Time.diff(timestamp1, timestamp2, :microseconds)
  end
end
```

### DIFF
--- a/lib/time/time.ex
+++ b/lib/time/time.ex
@@ -147,7 +147,7 @@ defmodule Timex.Time do
   @doc """
   Return a timestamp representing a time lapse of length 0.
 
-    Time.convert(Time.zero, :sec)
+    Time.convert(Time.zero, :secs)
     #=> 0
 
   Can be useful for operations on collections of timestamps. For instance,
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks Alexei! If you have a chance, could you open up a new issue as an RFC for making that change? Since it's backwards incompatible I'm a bit leery of doing so, but if there is enough agreement out there, I could plan on making that change prior to 1.0.

PR #36, opened 2015-01-16
Add moduledoc entry ala Dave Thomas

bbense:
Simple tweak to add moduledoc entry based on README.md ala Dave Thomas.


bitwalker:
Oh nice, I'll have to remember that trick!

PR #37, opened 2015-02-17
Add :strftime to parsing example

rorymckinley:
I think the :strftime argument of the one parsing example was missing. Unless you were using it as a n00b detector - in which case, you got me :)


bitwalker:
Oh good catch! Thanks for the PR!

rorymckinley:
Thanks - and thanks for an awesome library!

PR #40, opened 2015-03-20
Fix/simplify from/to_timestamp/convert/to_x conversions and codify conversion values

wishdev:
I apologize if I combined a couple of pull requests here......

Three things in this request, 

First - modify the Time.from and Time.to_timestamp functions to work properly. The msec variants did not properly multiple the msecs by 1000 so the values were not correct. The secs variants also did not work. The date_test modification clearly is an example of the problem - there are more then 63 seconds between Jan 1st 1971 and Jan 1st 1973.

Second - modified the functions so they, where possible, call down to a single version which does the work (either usecs or secs variant) and then convert the value back to the required type. I think the less places we're actually doing work the better.

Lastly - I also switched out the multiple locations that multiplied by an amount with an set of substitutions that declare what the value is.

Tests added and passed. If more tests desired/required please let me know...

### DIFF
--- a/lib/time/time.ex
+++ b/lib/time/time.ex
@@ -1,31 +1,41 @@
 defmodule Timex.Time do
 
+  @usecs_in_sec 1_000_000
+  @usecs_in_msec 1_000
+
+  @msecs_in_sec 1_000
+
+  @secs_in_min 60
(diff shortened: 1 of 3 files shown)


bitwalker:
This is great thanks!

Issue #48, opened 2015-06-10
milliseconds get lost when converting date to different timezone


hamann:
```elixir
iex(7)> "2015-06-10T10:11:57.315+0000" |> Timex.DateFormat.parse!("{ISO}") |> Timex.DateFormat.format!("{ISO}")
"2015-06-10T10:11:57.315+0000"
iex(8)> "2015-06-10T10:11:57.315+0000" |> Timex.DateFormat.parse!("{ISO}") |> Timex.Timezone.convert(Timex.Timezone.get("UTC")) |> Timex.DateFormat.format!("{ISO}")
"2015-06-10T10:11:57.315+0000"
iex(9)> "2015-06-10T10:11:57.315+0000" |> Timex.DateFormat.parse!("{ISO}") |> Timex.Timezone.convert(Timex.Timezone.get("CET")) |> Timex.DateFormat.format!("{ISO}")
"2015-06-10T12:11:57+0200" # <------
iex(10)> "2015-06-10T10:11:57.315+0200" |> Timex.DateFormat.parse!("{ISO}") |> Timex.Timezone.convert(Timex.Timezone.get("CET")) |> Timex.DateFormat.format!("{ISO}")
"2015-06-10T10:11:57.315+0200"


iex(12)> "2015-06-10T10:11:57.315+0000" |> Timex.DateFormat.parse!("{ISO}") |> Timex.Timezone.convert(Timex.Timezone.get("CET"))
%Timex.DateTime{calendar: :gregorian, day: 10, hour: 12, minute: 11, month: 6,
 ms: 0, second: 57,
 timezone: %Timex.TimezoneInfo{dst_abbreviation: "CEST",
  dst_end_day: {:last, :sun, :oct}, dst_end_time: {3, 0}, dst_name: "CEST",
  dst_start_day: {:last, :sun, :mar}, dst_start_time: {2, 0}, full_name: "CET",
  gmt_offset_dst: 60, gmt_offset_std: 60, standard_abbreviation: "CET",
  standard_name: "CET"}, year: 2015}
```

PR #49, opened 2015-06-10
 add milliseconds from source datetime struct when converting to new …

hamann:
fixes https://github.com/bitwalker/timex/issues/48


bitwalker:
Thanks!

PR #53, opened 2015-06-24
Add tzdata to :included_applications (for releases)

pma:


bitwalker:
It should go in `included_applications` actually, since it does not need to be started, only loaded. Could you make that change?

pma:
@bitwalker Changed to use :included_applications

bitwalker:
@pma Thanks!

Issue #68, opened 2015-07-26
Date.local is around 280 times slower in version 0.14.3 and 0.16.2 compared to 0.13.5


nicksanders:
Using the test below, mix test --trace shows Date.local is around 280 times slower in version 0.14.3 and 0.16.2 when compared to version 0.13.5

```elixir
test "test date now" do
  for _ <- 1..5000, do: Date.now
end

test "test date local" do
  for _ <- 1..5000, do: Date.local
end

test "test date local iso" do
  for _ <- 1..5000, do: DateFormat.format!(Date.local, "{ISO}")
end
```

**0.13.5**
```bash
[nick@zippy timetest]$ mix test --trace

TimetestTest
  * test date local (75.6ms)
  * test date local iso (800.2ms)
  * test date now (20.3ms)


Finished in 0.9 seconds (0.09s on load, 0.8s on tests)
3 tests, 0 failures
```

**0.14.3**
```bash
[nick@zippy timetest]$ mix test --trace

TimetestTest
  * test date now (31.0ms)
  * test date local iso (21997.8ms)
  * test date local (21219.1ms)


Finished in 43.3 seconds (0.08s on load, 43.2s on tests)
3 tests, 0 failures
```

**0.16.2**
```bash
[nick@zippy timetest]$ mix test --trace

TimetestTest
  * test date now (52.7ms)
  * test date local iso (23073.4ms)
  * test date local (22049.6ms)


Finished in 45.2 seconds (0.09s on load, 45.1s on tests)
3 tests, 0 failures
```

bitwalker:
Released in 0.17.0

PR #69, opened 2015-07-26
Cache local timezone to improve performance

nicksanders:
#68 

**Before**
5000 x Date.local takes 23073.4ms

**After**
5000 x Date.local takes 185.3ms

### DIFF
--- a/lib/timezone/timezone_local.ex
+++ b/lib/timezone/timezone_local.ex
@@ -44,7 +44,7 @@ defmodule Timex.Timezone.Local do
 
   def lookup(), do: Date.now |> lookup
   def lookup(%DateTime{} = date) do
-    tz = Application.get_env(:timex, :timezone)
+    tz = Application.get_env(:timex, :local_timezone)
     if tz == nil do
       tz = case :os.type() do
         {:unix, :darwin} -> localtz(:osx, date)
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -12,7 +12,7 @@ defmodule Timex.Mixfile do
   end
 
   def application do
-    [included_applications: [:tzdata]]
+    [included_applications: [:tzdata], timezone: nil]
```
bitwalker: Is this necessary to add? `get_env` and `put_env` should work fine without listing it here.

lib/timezone/timezone_local.ex
```diff
@@ -44,13 +44,18 @@ defmodule Timex.Timezone.Local do
 
   def lookup(), do: Date.now |> lookup
   def lookup(%DateTime{} = date) do
-    case :os.type() do
-      {:unix, :darwin} -> localtz(:osx, date)
-      {:unix, _}       -> localtz(:unix, date)
-      {:nt}            -> localtz(:win, date)
-      {:win32, :nt}    -> localtz(:win, date)
-      _                -> raise "Unsupported operating system!"
+    tz = Application.get_env(:timex, :timezone)
```
bitwalker: I think we should change it from `:timezone` to `:local_timezone` just to be specific as to what is being set here.

bitwalker:
Looks good! Just have a few comments to address and I can merge this :).

I'm sure I was caching this before, which is why I suspect it slowed down drastically after it was removed, I just can't recall the specific reasoning as to why.

nicksanders:
Looks like version 0.13.5 was using 

Process.get(:local_timezone) and Process.put(:local_timezone, tz)

is that better than Application.get_env/env_put?

bitwalker:
Application.get_env is a better place to put this honestly, the process dictionary isn't the ideal place for this after thinking about it some more. Thanks for making the changes so quickly!

PR #70, opened 2015-08-11
Some work on Elixir 1.1 compatibility

MSch:
We're moving development to Elixir master in anticipation of Elixir 1.1. I've made some changes to make timex at least compile with the current Elixir master. Unfortunately the Access protocol was used in at least the DateTime parser, so this is broken at the moment.


bitwalker:
@MSch I've manually merged some of your changes, and gone through to verify that all the potential Access derivations won't be a problem. I've pushed a new release 0.18.1 which is fully supported on Elixir 1.1.0-dev. Thanks for the heads up!

PR #76, opened 2015-09-08
use tzdata 0.5.2

sasa1977:
There was a bug in tzdata 0.5.1 which caused repeated errors and OTP app crash if the OTP release is not in the sam absolute path it was built in. See [here](https://github.com/lau/tzdata/pull/6) for more details.


bitwalker:
Thanks @sasa1977! I'll push a new release to hex ASAP.

PR #82, opened 2015-10-03
Add support for milliseconds in `Date.to_timestamp/2`

lafka:
Converting from `DateTime` to erlang timestamp looses sub-second precision

```
> Timex.DateFormat.parse!("2015-10-03T00:47:37.861234Z", "{ISO}") |> Timex.Date.to_timestamp
#> actual:   {1443, 833257, 0}
#> expected: {1443, 833257, 861234}
#> patched:  {1443, 833257, 861000}
```

As shown above, the patch does not account for the microsecond part (since `Timex.DateTime` doesn't) and rounds it to the nearest millisecond. It could be an idea to add a `:us` parameter to `DateTime` but that would need some discussion on best implementation for backward compatibility.

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -377,13 +377,13 @@ defmodule Timex.Date do
   @spec to_timestamp(DateTime.t, :epoch | :zero) :: timestamp
   def to_timestamp(date, reference \\ :epoch)
 
-  def to_timestamp(date, :epoch) do
+  def to_timestamp(%DateTime{:ms => ms} = date, :epoch) do
     sec = to_secs(date)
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks! I think adding microsecond support is eventually going to be necessary. I've been recommending using `Time` for that, but when parsing datetime strings with such precision like in your example, we shouldn't throw that away. It should be relatively straightforward to add a new `:us` field, but there are a number of places that would need to be updated to use the higher precision (for instance `to_secs` is used in a lot of places, which rounds to the nearest second). I should have some time soon to work on that :)

lafka:
Sounds good, let me know if you need some help.

Issue #89, opened 2015-10-12
RFC: beginning_of_... and end_of_... helpers


krishandley:
Would you be interested in adding any of the following helper functions from active support?

* beginning_of_year
* beginning_of_month
* beginning_of_week
* beginning_of_day
* end_of_year
* end_of_month
* end_of_week
* end_of_day

Example from the ruby world
http://apidock.com/rails/ActiveSupport/CoreExtensions/Date/Calculations/beginning_of_week

I was making a calendar and needed to get the beginning and end of the month, and week. I have written some code for this because I've been trying to learn Elixir, but opening an issue to get your thoughts.

bitwalker:
I'm open to adding these helpers, however let's break it up into multiple PRs and start with `beginning_of/end_of_year`. It'll be easier to review + merge, and the complexity increases as the timespan becomes more granular. Sound good?

krishandley:
Great, sounds good to me :)

krishandley:
Many thanks for accepting the helpers and a few elixir tips along the way :smile:

bitwalker:
No problem! Thanks for the contributions!

PR #91, opened 2015-10-12
Add Date.beginning_of_year and Date.end_of_year

krishandley:
Adds helper functions for setting the date to the beginning of and end of a year as discussed here #89

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -1386,4 +1386,64 @@ defmodule Timex.Date do
     end_of_month(from {year, 3 * quarter(month), 1})
   end
 
+  @doc """
+  Given a date or a number create a date at the beginning of that year
+
+  Examples
+
+      iex> date = #{__MODULE__}.from {{2015, 6, 15}, {0, 0, 0, 0}}
+      iex> #{__MODULE__}.beginning_of_year date
+      #{__MODULE__}.from {{2015, 1, 1}, {0, 0, 0, 0}}
+
+      iex> #{__MODULE__}.beginning_of_year 2015
+      #{__MODULE__}.from {{2015, 1, 1}, {0, 0, 0, 0}}
+
+      iex> #{__MODULE__}.beginning_of_year 2015, "Europe/London"
+      #{__MODULE__}.from {{2015, 1, 1}, {0, 0, 0, 0}}, "Europe/London"
+
+  """
+  @spec beginning_of_year(DateTime.t | year, String.t) :: DateTime.t
+  def beginning_of_year(%DateTime{year: year, timezone: tz}) when not is_nil(tz) do
+    beginning_of_year year, tz.full_name
+  end
+  def beginning_of_year(%DateTime{year: year}) do
+    beginning_of_year year
+  end
+  def beginning_of_year(year) do
+    from {year, 1, 1}
+  end
+  def beginning_of_year(year, timezone) do
+    from {year, 1, 1}, timezone
+  end
+
+  @doc """
+  Given a date or a number create a date at the end of that year
+
+  Examples
+
+      iex> date = #{__MODULE__}.from {{2015, 6, 15}, {0, 0, 0, 0}}
+      iex> #{__MODULE__}.end_of_year date
+      #{__MODULE__}.from {{2015, 12, 31}, {23, 59, 59}}
+
+      iex> #{__MODULE__}.end_of_year 2015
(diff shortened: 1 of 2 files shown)


lib/date/date.ex
```diff
@@ -1298,4 +1298,18 @@ defmodule Timex.Date do
     {d,{h,m,s,round(u/1000)}}
   end
 
+  @doc """
+  Set the date to the beginning of the year
```
bitwalker: How often do people want to provide a DateTime and get a completely different one back (in this case representing the beginning of the year)? It seems to me that these should be functions in the vein of `Date.iso_week`, etc., which produce a DateTime from some value. In this case, there is one required value (year), and one optional value (the timezone). These functions are more useful if they can be the starting point for getting a DateTime than require you to already have one, since if you already have a DateTime, it's trivial to extract the parts you need to call this function anyway. Thoughts?

lib/date/date.ex
```diff
@@ -1298,4 +1298,18 @@ defmodule Timex.Date do
...
+  """
+  def beginning_of_year(%DateTime{:year => year, :timezone => tz}) do
+    from {year, 1, 1}, tz.full_name
+  end
+
+  @doc """
+  Set the date to the end of the year
```
bitwalker: If possible, I'd like to use doctests for new API functions, because 1.) they provide better documentation for end users, and 2.) it is much easier to keep the tests and code in sync when they are next to each other. You can always provide both sets of tests (and I would encourage that too if you have functions with many possible variants/edge cases).

lib/date/date.ex
krishandley: Agreed and updated

krishandley:
What do the `@spec` lines do?

bitwalker:
@krishandley Oh damn, forgot to make a note of that, please do add those as well. Those are called typespecs, they are used by `dialyzer` to do static analysis of Elixir/Erlang code. You can use existing functions as references, but I'd do something like the following for both:

```elixir
@spec beginning_of_year(year, String.t) :: DateTime.t
```

In the above, `year` is already defined as a type at the top of `date.ex`, along with many other date/time related types.

bitwalker:
@krishandley Looks good, can you squash your commits and I'll merge?

tubedude:
Hey @krishandley  just curious. So in this implementation you give a integer representing the year and you get the date beginning or end of that year.
How do you expect to call it if you have a `Date` instead of a number? is it expected to find the Year of the Date and then pass that?
When we move to begining_of_month wouldn't it be more straightforward if the function accepted a `Date` instead?

krishandley:
@tubedude Yes that would be a good idea. I was also wondering if this might be better in a different module, something like Year.beginning_of, Month.beginning_of etc so the main Date module doesn't get cluttered

bitwalker:
@tubedude @krishandley You would simply call it like `Date.beginning_of_year(date.year)`. It's easy to get date properties from a DateTime, so I'd prefer to keep the the Date API as versatile as possible by not requiring a DateTime unless it's needed.

krishandley:
@bitwalker would you be opposed to allowing dates and years? Or would you prefer just the year?

tubedude:
@bitwalker When thinking about Year I think it looks good: `Date.beginning_of_year(date.year)`. But when we move to `Date.beginning_of_month(date.year, date.month)` I feel it's a bit odd. I'd rather have `Date.beginning_of_month(date)`.

bitwalker:
@krishandley @tubedude I'm open to having two versions of the same function, one which takes the DateTime and delegates to the one which takes the integer values for the required parameters. I just don't want functions that only take a DateTIme when the full DateTime is not required.

krishandley:
@bitwalker I've updated it to allow dates to be passed as well, but travis doesn't seem to want to run the Elixir 1.1 job :cry: Let me know if it's ok and I'll squash it.

bitwalker:
@krishandley Should be good to go, could you squash and then rebase your branch on master? I think the travis error is because your branch is based on an older version of timex and some changes in that file have been made since you forked. If that doesn't fix it, I'll merge anyway.

bitwalker:
@krishandley I just noticed there are no tests in this PR, could you add a few? Once there are a few which give good coverage, I'll merge this :)

krishandley:
The tests are in the docs like you asked for? But I can add more tests in the normal tests dir too.

bitwalker:
@krishandley /facepalm. No you're fine, some days man...

PR #93, opened 2015-10-19
Fix grammar in docs: “it's” → “its”

roryokane:
“It’s” is short for “it is”. “Its” is the possessive. I searched for all usages of each version and replaced the incorrect usages.

Also, in one place, I changed a comma to a semicolon.


bitwalker:
Thanks!

PR #95, opened 2015-11-06
Handle the :us token when using %f in strftime parsing

nickveys:
Tried parsing a custom format with sub-second precision and got an error: `Unrecognized token: us`.  Added support to the `update_date` function along w/a test.


bitwalker:
Thanks! This will be in a new release sometime today.

PR #105, opened 2015-11-24
Adding new Time exception and raising test coverage.

tubedude:
Skipping suggesting unit for Elixir under version 1.1 as it depends on the String.jaro_distance/1

### DIFF
--- a/lib/time/exceptions.ex
+++ b/lib/time/exceptions.ex
@@ -1,0 +1,29 @@
+defmodule Timex.Time.InvalidUnitError do
+  defexception [:unit, :message]
+
+  def exception(opts) do
+    unit = opts[:unit]
+    %Timex.Time.InvalidUnitError{unit: unit, message: msg(to_string unit)}
+  end
+
+  defp msg(unit) do
+    msg = "The unit #{unit} could not be found"
+    case Version.compare(System.version, "1.1.0") in [:gt, :eq] && did_you_mean(unit) do
+      {similar, score} when score > 0.8 ->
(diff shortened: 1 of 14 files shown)


lib/time/exceptions.ex
```diff
@@ -0,0 +1,35 @@
...
+		Enum.map(Timex.Time.all_units, &to_string/1)
+		|> Enum.reduce({nil, 0}, &max_similar(&1, unknown_unit, &2))
+	end
+
+	defp max_similar(source, target, {_, current} = best) do
+		score = String.jaro_distance(source, target)
+		if score < current, do: best, else: {source, score}
+	end
+
+	defp updated_elixir_version do
```
bitwalker: This can simply be:

```elixir
Version.compare(System.version, "1.1.0") in [:gt, :eq]
```

lib/time/exceptions.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Timex.Time.InvalidUnitError do
```
bitwalker: Looks good, but can you change the spacing to 2 spaces per tab instead of 4? Looks like only this file has tab-width of 4 for some reason.

tubedude:
@bitwalker I kinda messed up with the git by rebasing a commit I shouldn't. But it should not affect the end result, as the previous commit was already merged.

bitwalker:
I'm closing this for now, since v2 moves to a non-exception based error system, and the internals have been heavily refactored. If you'd like to refactor your changes to accomodate that, I'm still open to merging the better errors.

PR #106, opened 2015-11-25
Add a quick example of using strftime to format a DateTime

peburrows:
I didn't mind digging into the code to figure out how to do this, but it seems like something that should be mentioned in the README.


bitwalker:
@peburrows The format specification is detailed in `Timex.Format.DateTime.Formatters.Strftime`, but you're right, we don't have many examples of it's usage in the README or elsewhere. Thanks for the PR though, if you wanted to add a few more examples, feel free!

PR #110, opened 2015-12-04
Add Unix timestamp example to README.md

cwc:
Because it's a common use and I keep forgetting how to do it!


bitwalker:
Thanks for the PR!

PR #115, opened 2016-01-05
fix time diff

xbinxu:
current implementation of diff(t1, t2, :timestamp) has an error, it will raise arithmetic exception when mega == 1, then secs = div(microsecs, 0)


bitwalker:
Thanks for the PR! Could you add tests for the failing case + cases which show the diff implementation works as expected?

xbinxu:
@bitwalker, test case added, :)

PR #119, opened 2016-02-10
Readme: use the :timex app instead of :tzdata

rstacruz:
Elixir noob here. Doesn't this make more sense?


bitwalker:
Thanks! I added it to the Getting Started docs, but missed this one. Cheers!

PR #121, opened 2016-02-11
Some small optimizations and removal of imperative if uses.

TBK145:
In Elixir 1.3 imperative uses of if are going to emit warnings, so I took a look at this library and found some ways to remove them and optimize it. This PR changes 3 things.
- In Date.validate/1, instead of checking if day > max_day, just return min(day, max_day).
- Because of how div/2 and rem/2 behave, there is no need to check for >= @million in Time.normalize/1.
- In Timezone.Local.lookup/1, instead of checking if tz == nil, it's now one big case statement, removing the last imperative if.

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -1293,10 +1293,7 @@ defmodule Timex.Date do
   defp validate({year, month, day}) do
     # Check if we got past the last day of the month
     max_day = days_in_month(year, month)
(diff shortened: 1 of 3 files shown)


bitwalker:
Awesome! I've been meaning to go back and fix instances of imperative binding like that, thanks for taking the time to do so!

PR #122, opened 2016-02-19
Add Date.from ability to handle Phoenix datetime_select changeset

mgnshawn:
I had added a Date.from version that takes the params from a Phoenix Edit View's datetime_select control and returns a Timex.Datetime struct to allow the creation of a changeset.
Example Usage:

```elixir
def update(conn, %{"id" => id, "cold_lead" => cold_lead_params}) do
  if(cold_lead_params||0 != 0) do
    if(Map.has_key?(cold_lead_params,"discovered")) do
      unless(Map.has_key?(cold_lead_params["discovered"],:calendar)) do
        cold_lead_params = update_in(cold_lead_params["discovered"], &(Timex.Date.from &1))
      end
    end
  end
  cold_lead = Repo.get!(ColdLead, id)
  changeset = ColdLead.changeset(cold_lead, cold_lead_params)
```

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -284,6 +284,7 @@ defmodule Timex.Date do
   ## Examples
 
       > Date.from(:erlang.universaltime)             #=> %DateTime{...}
+      > Date.from(datetime_select params)            #=> %DateTime{...}
       > Date.from(:erlang.localtime)                 #=> %Datetime{...}
       > Date.from(:erlang.localtime, :local)         #=> %DateTime{...}
       > Date.from({2014,3,16}, "America/Chicago")    #=> %DateTime{...}
@@ -297,6 +298,9 @@ defmodule Timex.Date do
 
   def from({y,m,d} = date, :utc) when is_integer(y) and is_integer(m) and is_integer(d),
     do: construct({date, {0,0,0}}, %TimezoneInfo{})
+  def from(%{"day" => _, "hour" => _, "min" => _, "month" => _, "year" => _} = dtm, :utc) do
+    construct({{String.to_integer(dtm["year"]),String.to_integer(dtm["month"]),String.to_integer(dtm["day"])},{String.to_integer(dtm["hour"]),String.to_integer(dtm["min"]),00}}, %TimezoneInfo{})
+  end
   def from({{_,_,_},{_,_,_}} = datetime, :utc),
     do: construct(datetime, %TimezoneInfo{})
   def from({{_,_,_},{_,_,_,_}} = datetime, :utc),
--- a/test/date_test.exs
+++ b/test/date_test.exs
@@ -138,6 +138,11 @@ defmodule DateTests do
     assert %DateTime{year: 1970, month: 2, day: 1, hour: 0, minute: 0, second: 0} = Date.from(31, :days)
   end
 
+  test "from datetime_select" do
+    date = %{"day" => "05", "hour" => "12", "min" => "35", "month" => "2", "year" => "2016"}
+    assert %DateTime{year: 2016, month: 2, day: 05, hour: 12, minute: 35, second: 0} = date |> Date.from
+  end
+
   test "to timestamp" do
     assert {0,0,0} === Date.epoch |> Date.to_timestamp
     assert {62167,219200,0} === Date.epoch |> Date.to_timestamp(:zero)


bitwalker:
@mgnshawn Thanks for the PR! I'm thinking to merge this manually in order to add the typedef for the datetime_select_params type, flesh out the docs a bit more, and reorder the clause after the tuple-type clauses. I'll merge the changes with you set as the author of the commits, but before I do that I want to make sure that's ok with you, or if you'd like to make the changes yourself to get more familiar with the typical PR-workflow I can add comments to the diff for you to review.

Also, as a completely random aside, your example code would be more idiomatic Elixir/readable if you wrote it like so:

```elixir
def update(conn, %{"id" => id, "cold_lead" => %{"discovered" => %{calendar: _} = discovered} = params) do
  params = %{params | "discovered" => Timex.Date.from(discovered)}
  Repo.get!(ColdLead, id) |> ColdLead.changeset(params)
end
def update(conn, %{"id" => id, "cold_lead" => params}) when is_map(params) do
  Repo.get!(ColdLead, id) |> ColdLead.changeset(params)
end
def update(conn, %{"id" => id) do
  Repo.get!(ColdLead, id) |> ColdLead.changeset(%{})
end
```

By using pattern matching in the function head, you can break the different branches into smaller bits,
makes it a lot easier to find the branch you care about, and is overall a lot easier to read. In future versions of Elixir, the imperative assignment of `cold_lead_params` in your original example will not compile, so you'd have to change it anyway, but since I'm assuming you're relatively new to Elixir, I thought it might be useful to see the more idiomatic approach :)

Thanks again, and let me know what you'd like me to do as far as merging!

mgnshawn:
Paul,
Please do the merge manually so I can learn from how a more seasoned elixir
developer would approach it.  I'm only a week into Elixir via (Programming
Elixir and Programming Phoenix Books which I burned through), coming from a
C#.Net, PHP, and rough RoR skill set.  Preferred in that ordered.  I am
very excited about Elixir though and it's performance on concurrent
 sockets!

I do prefer the pattern matching head as you've created, but could you
clarify for me:

*Could you expand on how ""%{"discovered" => %{calendar: _} =
discovered} = params"" operates, in terms of "if value for key
discovered has a struct with a key of calendar, set to discovered,
else set params".  I haven't come across this type of statement yet
but it looks efficient.

def update(conn, %{"id" => id, "cold_lead" => %{"discovered" =>
%{calendar: _} = discovered} = params) do


*Is this supposed to be a double pipe, if not could you explain what a
single pipe does?
  params = %{params | "discovered" => Timex.Date.from(discovered)}


Thank you,

-Shawn


On Fri, Feb 19, 2016 at 12:35 PM, Paul Schoenfelder <
notifications@github.com> wrote:

> @mgnshawn <https://github.com/mgnshawn> Thanks for the PR! I'm thinking
> to merge this manually in order to add the typedef for the
> datetime_select_params type, flesh out the docs a bit more, and reorder the
> clause after the tuple-type clauses. I'll merge the changes with you set as
> the author of the commits, but before I do that I want to make sure that's
> ok with you, or if you'd like to make the changes yourself to get more
> familiar with the typical PR-workflow I can add comments to the diff for
> you to review.
>
> Also, as a completely random aside, your example code would be more
> idiomatic Elixir/readable if you wrote it like so:
>
> def update(conn, %{"id" => id, "cold_lead" => %{"discovered" => %{calendar: _} = discovered} = params) do
>   params = %{params | "discovered" => Timex.Date.from(discovered)}
>   Repo.get!(ColdLead, id) |> ColdLead.changeset(params)enddef update(conn, %{"id" => id, "cold_lead" => params}) when is_map(params) do
>   Repo.get!(ColdLead, id) |> ColdLead.changeset(params)enddef update(conn, %{"id" => id) do
>   Repo.get!(ColdLead, id) |> ColdLead.changeset(%{})end
>
> By using pattern matching in the function head, you can break the
> different branches into smaller bits,
> makes it a lot easier to find the branch you care about, and is overall a
> lot easier to read. In future versions of Elixir, the imperative assignment
> of cold_lead_params in your original example will not compile, so you'd
> have to change it anyway, but since I'm assuming you're relatively new to
> Elixir, I thought it might be useful to see the more idiomatic approach :)
>
> Thanks again, and let me know what you'd like me to do as far as merging!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bitwalker/timex/pull/122#issuecomment-186319123>.
>



-- 

Shawn Cook

Director of IT

712 Ferry Street

Easton, PA 18042

Phone: 508-921-4045

Fax: 610-252-3331

bitwalker:
I came from a heavy C# background myself, so I can relate :). I'll get this merged today, and to answer your questions:

> *Could you expand on how ""%{"discovered" => %{calendar: _} =
discovered} = params"" operates, in terms of "if value for key
discovered has a struct with a key of calendar, set to discovered,
else set params".  I haven't come across this type of statement yet
but it looks efficient.

> *Is this supposed to be a double pipe, if not could you explain what a
single pipe does?

A couple things that might clarify the meaning:

- The pattern match isn't for a struct per se (which would look like `%SomeStruct{:some_key => "some value"}`), just a map. Since structs are ultimately maps, you can match on both using the same patterns (the only key difference being that struct keys will always be atoms, but map keys can be any Elixir/Erlang term).
- There are two ways to match/create maps:
  - `%{key: value}`
  - `%{:key => value}`
  - both of which are equivalent to `Map.put(%{}, :key, value)` or `put_in(%{}, [:key], value)`
- You can also build a new map from an old map like so:
  - `%{old_map | :key => new_value}`
  - which is equivalent to `Map.put(old_map, :key, new_value)`
  - or `put_in(old_map, [:key], new_value)`
- To put the first clause in the form of a list of steps describing the match, it would be:
  - any term for the first parameter is a match, but only if:
  - the second parameter is a map, and:
  - it has at least two string keys named `id` and `cold_lead`, and:
  - the `cold_lead` value is also a map, with at least one string key named `discovered`, and:
  - `discovered` is a map with at least one atom key named `calendar` which can have any value, because it is ignored (`_` is both a wildcard match and no binding will be made)
  - the value of `discovered` will be bound to the name `discovered`
  - the value of the second parameter will be bound to the name `params`
- If a function clause does not match it's parameters, the next clause (in source order) will be tried, until a match is found, or in the case of no matching clauses, a FunctionClauseError is raised.

bitwalker:
@mgnshawn See the commit referenced above to see the diff, the only significant changes were to the `from/2` implementation. `String.to_integer` raises an error if it fails, where other clauses of `from/2` return an error tuple, so my changes fold over the provided map, converting each of the string keys to an integer using `Integer.parse`, and if an invalid value is provided, the result will be `:error` instead of a map of strings -> integers. I also added in handling for timezones other than `:utc`, not that it will ever be used in your case, but it matches the behaviour of the other clauses now.

PR #123, opened 2016-02-24
Disable tzdata's auto-update during compilation

alco:


lau:
The problem is Tzdata needing to run during compilation of Timex. Why is Tzdata running during compilation? Are macros using tzdata using compilation? If so, it would be ideal to get rid of that. Tzdata should only be used during runtime. Otherwise you risk some data changing in Tzdata, but the old data is compiled into Timex.

bitwalker:
@lau Yes, the current behaviour is present due to legacy support for UNIX-formatted datetime strings, which use 3-letter abbreviations such as PST. I have an alternate path for supporting these in progress, but when I transitioned to tzdata I used this method as an intermediate (and imperfect solution).

lau:
I see. Good to hear that using the data at compile time is just an intermediate solution.

bitwalker:
@alco @lau - just in case you are interested, I implemented the proper solution for handling POSIX timezone abbreviations in master. I plan to extend support for handling of POSIX timezone abbreviations to include those not defined in the Olson database, but this covers the most common use-case. This also removes any compile-time dependency on tzdata entirely.

alco:
:+1: Thanks for such a quick turnaround!

PR #140, opened 2016-03-26
Fixing Timex.parse typespecs

nickveys:
Timex.parse typespecs currently say it returns Strings, updating to DateTime.t


bitwalker:
Nice catch! Thanks for the PR!

PR #143, opened 2016-03-30
add :tzdata to applications list

schaary:
When you mention to add `:tzdata` to the applications list, then it is confusing, when you don't do it in the example code. I've add `:tzdata` to the applications list in the example code snippets in the `README.md` and the `Getting Started.md` files

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Timex
+#accessed# Timex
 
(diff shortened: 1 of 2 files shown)


bitwalker:
Timex actually does this for you, you don't have to add it yourself. Would you be open to changing this to instead remove the language around adding `:tzdata`?

PR #145, opened 2016-03-31
fix Timex.diff for equal datetimes and timestamp granularity

ronnieoverby:
### Summary of changes

In my own app I was diff'ing 2 datetimes, asking for a timestamp back. In some cases I would diff 2 equal datetimes, and get a zero returned instead of `{0,0,0}` even though I requested a timestamp.

My PR fixes this, which I believe to be a bug.

### Checklist

Prior to this fix, the :timestamp type was ignored and 0 integer was returned


nickveys:
Hah, nice, I literally just ran into this and was about to create a PR.  Thanks for fixing this!  Hope it gets merged quick!

bitwalker:
Oh nice catch! I'll get this merged and released ASAP

PR #146, opened 2016-04-01
Fix parsing for ISO-822 and ISO-1123 formatted dates

spscream:
### Summary of changes
When I tried to parse Date header from email message with the following format:
"Fri, 1 Apr 2016 20:29:18 +0300"
I got the following exception:
```
    ** (exit) an exception was raised:
    ** (Timex.Parse.ParseError) Expected `day of month` at line 1, column 6.
        (timex) lib/parse/datetime/parser.ex:87: Timex.Parse.DateTime.Parser.parse!/3
```

Accordingly [RFC-1123](https://www.ietf.org/rfc/rfc1123.txt) (Section 5.2.14) syntax for date is 
```
            date = 1*2DIGIT month 2*4DIGIT
```
For [RFC-822](https://www.w3.org/Protocols/rfc822/#z28):
```
date =  1*2DIGIT month 2DIGIT  
```
I changed parsers and added additional test cases for this behavior.

### Checklist

- [x ] New functions have typespecs, changed functions were updated
- [ x] Same for documentation, including moduledocs
- [ x] Tests were added or updated to cover changes
- [ x] Commits were squashed into a single coherent commit

### DIFF
--- a/lib/parse/datetime/parsers.ex
+++ b/lib/parse/datetime/parsers.ex
@@ -359,7 +359,7 @@ defmodule Timex.Parse.DateTime.Parsers do
     parts = [
       weekday_short(opts),
       literal(string(", ")),
-      day_of_month([padding: :zeroes, min: 2, max: 2]),
+      day_of_month([padding: :zeroes, min: 1, max: 2]),
       literal(space),
       month_short(opts),
       literal(space),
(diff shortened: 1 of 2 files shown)


bitwalker:
Hey thanks for fixing this, I think I borrowed the month/day parsers from ISO-8601 parser, and forgot to change the padding options :/

:100:

PR #147, opened 2016-04-01
add ASN.1 GeneralizedTime format capability

jeffweiss:
### Summary of changes

Prior to this commit Timex would, unstandably, barf if one attempted to
parse an ASN.1 GeneralizedTime format, which is the underlying format
for things like LDAP. This commit adds the ability to parse ASN.1
GeneralizedTime (UTC only).

### DIFF
--- a/lib/format/datetime/formatter.ex
+++ b/lib/format/datetime/formatter.ex
@@ -358,6 +358,29 @@ defmodule Timex.Format.DateTime.Formatter do
     sec   = format_token(locale, :sec, date, modifiers, flags, width_spec(2..2))
     "#{year}#{month}#{day}#{hour}#{min}#{sec}Z"
   end
+  def format_token(locale, :asn1_generalized_time, %DateTime{} = date, modifiers, _flags, _width) do
+    # `130305232519`
+    flags = [padding: :zeroes]
+    year  = format_token(locale, :year4, date, modifiers, flags, width_spec(4..4))
+    month = format_token(locale, :month, date, modifiers, flags, width_spec(2..2))
+    day   = format_token(locale, :day, date, modifiers, flags, width_spec(2..2))
+    hour  = format_token(locale, :hour24, date, modifiers, flags, width_spec(2..2))
+    min   = format_token(locale, :min, date, modifiers, flags, width_spec(2..2))
+    sec   = format_token(locale, :sec, date, modifiers, flags, width_spec(2..2))
+    "#{year}#{month}#{day}#{hour}#{min}#{sec}"
+  end
+  def format_token(locale, :asn1_generalized_time_z, %DateTime{} = date, modifiers, flags, width) do
+    # `130305232519Z`
+    date = Timezone.convert(date, "UTC")
+    base = format_token(locale, :asn1_generalized_time, date, modifiers, flags, width)
+    base <> "Z"
+  end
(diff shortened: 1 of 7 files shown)


bitwalker:
Hey @jeffweiss, this looks great! Would you be willing to also add the non-zulu variant as well? Seems like we should support both if we're going to support one of them. Should be able to apply the same logic as the ISO-8601 vs ISO-8601z parsers.

jeffweiss:
@bitwalker Digging in further, ASN.1 Generalized Time has 3 variants:
* ambiguous time zone, assumed local
* UTC
* local tz as offset from UTC

I'm realizing now that I'm not entirely certain what to **name** these variants. With just two it's fairly straightforward (though quite verbose: ASN1:GeneralizedTime[z]). But with 3, idk. What do you think about:
* ASN1:GeneralizedTime - `YYYYMMDDHH[MM[SS[.fff]]]`
* ASN1:GeneralizedTime:Z - `YYYYMMDDHH[MM[SS[.fff]]]Z`
* ASN1:GeneralizedTime:Extended - `YYYYMMDDHH[MM[SS[.fff]]]+-HHMM`

bitwalker:
How about GeneralizedTime[Z|TZ|]? I know it's about verbose to name things this way, but the advantage is clarity where used.

jeffweiss:
@bitwalker This latest amendment has `{ASN1:GeneralizedTime[:Z|:TZ]}`, to at least provide some consistency between this and the existing `{ASN1:UTCtime}`. If you'd rather, I can totally change to just `{GeneralizedTime[Z|TZ]}`. This ASN.1 format is far less prevalent than, say, RFC3339 or ISO8601, so I'm personally hesitant to just call it `GeneralizedTime` because it might encourage people to use it where RFC3339 would be a better (interop) choice.

jeffweiss:
oh, nevermind... :smile:

bitwalker:
Haha, damn, just missed your comment. Well it's fine to have all three I think. If someone is looking for the ASN.1 variants, they probably know which variant they are targeting. That said, if you think there is no need for the GeneralizedTime variant, I'd be willing to merge a PR which removes it. This is one of those times where I wish there was a way to get metrics on what APIs are being used - a dream that will never be realized probably :(

jeffweiss:
Haha, I think we're saying the same thing: People who know they need ASN.1 will go right to it, people who don't know will likely be intimidated by the ASN.1, thereby staying away. I think we're golden here.

That's an interesting thought about telemetry of which APIs are used. We could probably do it with statsd.  Small statsd server in AWS or the like, config parameter to turn on and off sending the UDP statsd packets.

bitwalker:
Definitely doable, but I know personally I would be really turned off by the idea of a library sending stats anywhere, even if I know it's for a good reason. I imagine most people would disable it, or even move to another library if they thought timex did that, and thus it would be pretty useless for making the kind of decisions I'd use it for. 

That said, it would in general be a really cool feature of the VM if one could turn on metrics for a given application or module and then be able to dump it to a file. So I could say "hey community, if you are interested in helping improve timex, do a dump of the timex calls from your apps and send them to me", and then run some analysis on those dumps to build a picture of what the most common use cases seem to be.

Best I can do now is look at the code and be like "that probably doesn't get used much, I'm going to deprecate it", and then wait for the issues to come in ;)

PR #149, opened 2016-04-02
collapse directives to 3 basic cases

jeffweiss:
### Summary of changes

### Checklist

Prior to this commit the directive were extremely repetitive without a
great deal of value and, in some cases, hiding the relationship between
the Directive type and the related Parser function. This commit distills
the directives into 3 basic cases:
* Directive type == Parser function
* Directive type != Parser function
* Directive type != Parser function, add zulu flag

Although none currently exist, if we encounter a case that does not fit
any of these 3 scenarios, then we can add it as a special case.

@bitwalker, if you feel like this doesn't make it easier to read or maintain, feel free to close/wont-merge. This made it easier for _me_ to reason about it, but I'm not sure it's a universal improvement.

### DIFF
--- a/lib/parse/datetime/tokenizers/directive.ex
+++ b/lib/parse/datetime/tokenizers/directive.ex
@@ -33,135 +33,41 @@ defmodule Timex.Parse.DateTime.Tokenizers.Directive do
     get(type, directive, flags, modifiers, width)
   end
 
-  # Years
-  def get(:year4, directive, flags, mods, width),
-    do: %Directive{type: :year4, value: directive, flags: flags, modifiers: mods, width: width, parser: Parsers.year4(flags)}
-  def get(:year2, directive, flags, mods, width),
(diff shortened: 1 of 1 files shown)


jeffweiss:
@bitwalker it looks like coverage only decreased the 90 line reduction in file size. :-/

jeffweiss:
I'll need to rebase after the merge of #147 :smiley: standby...

bitwalker:
Actually I really like this, it bothered me a bit that there was so much duplication there, but never enough where I took the time to refactor it. I think the coverage change is due to the fact that the uncovered lines now represent a greater percentage of the total lines of code, which I think is what you were saying, but yeah, not an issue.

Let me know when you want me to go ahead and merge :)

jeffweiss:
@bitwalker go for it :+1:

bitwalker:
Thanks!

PR #150, opened 2016-04-02
add support for partial-hour offset timezones

jeffweiss:
### Summary of changes

### Checklist

Prior to this commit Timex would produce a blantantly incorrect value
for parsing a timestamp with a partial-hour offset timezone, such as
that of Nepal (+05:45). The resulting timestamp would have the correct
wall time, but a truncated timezone (+05:00), which would result in
incorrect timestamps in UTC. This commit adds support for partial-hour
offset timezones by calling them out as special cases when attempting to
execute `Timex.Timezone.name_of`, which eventually gets invoked by
`Timex.parse`.

### DIFF
--- a/lib/timezone/timezone.ex
+++ b/lib/timezone/timezone.ex
@@ -77,6 +77,19 @@ defmodule Timex.Timezone do
   def name_of("Y"),    do: name_of(-12)
   def name_of("Z"),    do: "UTC"
   def name_of("UT"),   do: "UTC"
+  def name_of("+0330"), do: "Asia/Tehran"
+  def name_of("+0430"), do: "Asia/Kabul"
+  def name_of("+0530"), do: "Asia/Colombo"
+  def name_of("+0545"), do: "Asia/Kathmandu"
+  def name_of("+0630"), do: "Asia/Rangoon"
+  def name_of("+0830"), do: "Asia/Pyongyang"
+  def name_of("+0845"), do: "Australia/Eucla"
+  def name_of("+0930"), do: "Australia/Darwin"
+  def name_of("+1245"), do: "Pacific/Chatham"
+  def name_of("+1345"), do: "Pacific/Chatham"
+  def name_of("-0430"), do: "America/Caracas"
+  def name_of("-0330"), do: "America/St_Johns"
+  def name_of("-0230"), do: "America/St_Johns"
   def name_of(offset) when is_integer(offset) do
     if offset > 0 do
       "Etc/GMT-#{offset}"
--- a/test/timezone_test.exs
+++ b/test/timezone_test.exs
@@ -127,4 +127,25 @@ defmodule TimezoneTests do
 
     assert 63645015540 = DateTime.to_seconds(Timex.datetime(datetime, "Europe/Vienna"), :zero, [utc: true])
   end
+
+  test "name_of" do
+    assert {:error, {:invalid_timezone, :wat}} = Timezone.name_of(:wat)
+    assert {:error, {:invalid_timezone, "America/Not_A_Real_Timezone"}} = Timezone.name_of("America/Not_A_Real_Timezone")
+    assert "America/Los_Angeles" = Timezone.name_of("America/Los_Angeles")
+    assert "Etc/GMT+0500" = Timezone.name_of("GMT+0500")
+    assert "Etc/GMT-0500" = Timezone.name_of("GMT-0500")
+  end
+
+  test "partial-hour offset timezones" do
+    zones = [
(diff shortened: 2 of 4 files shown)


lib/timezone/timezone.ex
```diff
@@ -77,6 +77,19 @@ defmodule Timex.Timezone do
   def name_of("Y"),    do: name_of(-12)
   def name_of("Z"),    do: "UTC"
   def name_of("UT"),   do: "UTC"
+  def name_of("+0330"), do: "Asia/Tehran"
```
bitwalker: Yeah we definitely can't go this route, I would instead change this to return a variant of `Etc/GMT`, such as `Etc/GMT+5:45`, and modify `get` to special-case these and return a manually constructed `TimezoneInfo` struct.

For example, `Etc/GMT+5` from tzdata is:

```elixir
%{from: %{standard: :min, utc: :min, wall: :min}, std_off: 0,
    until: %{standard: :max, utc: :max, wall: :max}, utc_off: -18000,
    zone_abbr: "GMT+5"}
```

The same thing for `Etc/GMT+5:45` would be:

```elixir
%{from: %{standard: :min, utc: :min, wall: :min}, std_off: 0,
    until: %{standard: :max, utc: :max, wall: :max}, utc_off: -20700,
    zone_abbr: "GMT+5:45"}
```

And as a TimezoneInfo:

```elixir
"%TimezoneInfo{
  abbreviation: "GMT+5:45", 
  from: :min, 
  full_name: "Etc/GMT+5:45", 
  offset_std: 0, 
  offset_utc: -345, 
  until: :max}
```

This is the really the only valid way to parse these offsets, because we can't assume what zone was intended, only that the zone which was formatted had some offset - with date/times it's actually worse to assume more information than was given, and assuming the zone from an offset, even if a given offset is currently unique, is assuming too much.  At any time a new zone could be added, or an existing one changed, to use an odd offset like this. We can't ever make an assumption about what zone was meant (even if we can determine exactly what zone is referenced because it's the only one with such an offset at that time).

The alternative is to have some special handling in the timezone offset parsers which use the date/time context to see if it can locate the specific zone given the offset, but I dislike that because it's not consistent with how offsets have to be parsed in the general case (i.e. parsed into one of the etcetera zones, like `Etc/GMT+5`). Since the Olson timezone database does not contain any such etcetera zones, we have to roll it ourselves for cases like this.

Creating an etcetera zone for an odd offset is fine generally anyway when parsing datetimes and shifting them to another zone, since when shifting to another zone we only care about the current offset, not the rules of the zone, which only come into play when shifting through time - and combining parsing with shifts through time is a rare case, if done at all - and something I'd rather be able to say "this is how it behaves when you do that, for every offset", than to say "well sometimes we assume a zone based on the offset, and sometimes we just use the offset".

I think I've probably gone on too long, but wanted to make my justification clear :)

test/timezone_test.exs
```diff
@@ -127,4 +127,25 @@ defmodule TimezoneTests do
 
     assert 63645015540 = DateTime.to_seconds(Timex.datetime(datetime, "Europe/Vienna"), :zero, [utc: true])
   end
+
+  test "name_of" do
+    assert {:error, {:invalid_timezone, :wat}} = Timezone.name_of(:wat)
+    assert {:error, {:invalid_timezone, "America/Not_A_Real_Timezone"}} = Timezone.name_of("America/Not_A_Real_Timezone")
+    assert "America/Los_Angeles" = Timezone.name_of("America/Los_Angeles")
+    assert "Etc/GMT+0500" = Timezone.name_of("GMT+0500")
```
bitwalker: It's important to note that the etcetera zones have a specific format (non zero-padded), and that none of them support seconds granularity. If as I suggested we fake them, we should adhere to the convention, so this zone would be `Etc/GMT+5:45`. I want to be clear about the differentiation between hours and seconds, hence the colon.

lib/timezone/timezone.ex
jeffweiss: Cool. I have zero urgency around this. I knew the odd offsets existed and that's it. 

I'll try to update in the next couple evenings or so.

bitwalker: Yeah I totally understand :), you've been on fire the last few days anyway haha.

This particular issue is something I've thought about a few times, and I have mixed feelings. On the one hand, I want to avoid supporting edge cases not handled by the Olson database, on the other, I want to be pragmatic about supporting real world scenarios. I was mostly waiting for exactly this situation, someone encounters it and requests a fix. You took it a step further and actually implemented a fix, which is awesome! Since we're here though, might as well take it all the way :)

bitwalker:
Support for this is now in master. I haven't tested it heavily, so there may be places which need to be updated to handle it properly, but please give it a try and open a new issue if it is not working!

PR #153, opened 2016-04-04
diff of 7 days should == diff of 1 week

frpaulas:
### Summary of changes

I think when Timex.diff(a, b, :days) == 7
Timex.diff(a, b, :weeks) should == 1
currently Timex.diff(a, b, :weeks) == 0 when Timex(a, b, :days) == 7

### Checklist

d1 = Timex.date {2016, 3,28}; d2 = Timex.date {2016, 4,4}
Timex.diff d1, d2, :days # => 7
Timex.diff d1, d2, :weeks # => 1

### DIFF
--- a/lib/comparable/datetime.ex
+++ b/lib/comparable/datetime.ex
@@ -122,7 +122,7 @@ defimpl Timex.Comparable, for: Timex.DateTime do
     extra_days = rem(days, 7)
     actual_weeks = (if extra_days == 0, do: weeks, else: weeks + 1)
     cond do
-      actual_weeks == 1 && extra_days < 7 -> 0
+      actual_weeks == 1 && days < 7 -> 0
       :else -> actual_weeks
     end
   end
--- a/test/timex_test.exs
+++ b/test/timex_test.exs
@@ -239,6 +239,13 @@ defmodule TimexTests do
     date2 = Timex.datetime({1969,2,11})
     assert Timex.diff(date1, date2, :months) === 25
     assert Timex.diff(date2, date1, :months) === 25
+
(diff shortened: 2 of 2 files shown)


lib/comparable/datetime.ex
```diff
@@ -122,7 +122,7 @@ defimpl Timex.Comparable, for: Timex.DateTime do
     extra_days = rem(days, 7)
     actual_weeks = (if extra_days == 0, do: weeks, else: weeks + 1)
     cond do
-      actual_weeks == 1 && extra_days < 7 -> 0
+      actual_weeks == 1 && days < 7 -> 0
```
bitwalker: This has no effect (if actual_weeks is 1, days will always be greater than or equal to 7), instead, we should make the first branch of the cond `actual_weeks == 0 && extra_days < 7 -> 0`.

bitwalker:
Left a note, but once that's addressed I can merge. Thanks for the PR!

frpaulas:
Perhaps the first cond should be `days < 7 -> 0`. All tests pass, but when the 1st condition is 

`actual_weeks == 0 && extra_days < 7 -> 0`

testing {2016, 1, 1} & {2015, 12, 31} tests fails with 

> 1) test diff (TimexTests)
     test/timex_test.exs:187
     match (=) failed
     code: 0 = Timex.diff(date5, date6, :weeks)
     rhs:  1
     stacktrace:
       test/timex_test.exs:218

And inside `defp do_diff(adate, a, bdate, b, :weeks)`  I see
- days == 1
- weeks == 0
- extra_days == 1
- actual_weeks == 1

bitwalker:
Closing this since it has been fixed in master

PR #154, opened 2016-04-05
add docs and more explicit tests for fractional seconds

gaxunil:
### Summary of changes

So I needed to parse a date like this:
"2016-04-01T10:43:47.687"

the {ss} fractional seconds seemed to be missing from the docs, so I added that, then a couple test cases to show examples explicitly in format strings.

I tried to parse also via the strftime formats, but I just couldn't get it.
The only way I could get it to work is to extend out the factional to explicit microseconds in the date string:
"2016-04-01T10:43:47.687000"
And then parse with this format:
%Y-%m-%dT%H:%M:%S.%f"

Should "2016-04-01T10:43:47.687" be directly parseable via sttftime format?

Thanks for this great library!

### Checklist

- [n/a] New functions have typespecs, changed functions were updated
- [x ] Same for documentation, including moduledocs

### DIFF
--- a/lib/format/datetime/formatters/default.ex
+++ b/lib/format/datetime/formatters/default.ex
@@ -56,6 +56,7 @@ defmodule Timex.Format.DateTime.Formatters.Default do
   * `{h12}`     - hour of the day (1..12)
   * `{m}`       - minutes of the hour (00..59)
   * `{s}`       - seconds of the minute (00..60)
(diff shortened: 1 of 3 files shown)


bitwalker:
Thanks for the PR!

The strftime format is more explicit about how it parses things, there is unfortunately no directive for strftime which will parse both fractional and non-fractional seconds. We could probably add one of course, but the current implementation attempts to stay in parity with [this reference](linux.die.net/man/3/strftime).

PR #162, opened 2016-04-26
feat(translation): add german translation

schurig:
### Summary of changes

* added german translation

### Checklist


bitwalker:
This is great! Thanks so much!

PR #163, opened 2016-04-27
Fix shifting by milliseconds with current msecs value

khamilowicz:
### Summary of changes

Previous fix by milliseconds did not add current value of msecs.

### Checklist

Timex datetime that was already shifted by milliseconds should
use them when shifting time


bitwalker:
Thanks for this, and sorry for the delay!

PR #164, opened 2016-04-27
fix(german): fix and improve german translation

schurig:
### Summary of changes

* fixed encoding to UTF-8
* fixed typos

### Checklist


bitwalker:
Thanks!

PR #165, opened 2016-05-02
BUGFIX: shifting months to the end of a year would have resulted in a…

gutschilla:
### Summary of changes

This PR addresses issue 158 submitted by @mindreframer 

## Issue description

- when shifting dates by months and months > 12; month was set to rem( months, 12) 
  - which could result in setting month to zero which leads to dates like {2018,0,25} 
  - that dates consequently fail when used in other functions

## Bugfix description
- added a simple check for this and setting month to 12 and decrement year by 1 in this case
- added small test that addresses the bug

### DIFF
--- a/lib/datetime/datetime.ex
+++ b/lib/datetime/datetime.ex
@@ -792,6 +792,12 @@ defmodule Timex.DateTime do
       m < 0  -> %{datetime | :year => year + div(m, 12), :month => 12 + rem(m, 12)}
       :else  -> %{datetime | :month => m}
     end
+
+    # setting months to remainders may result in month = 0
+    if shifted.month == 0 do
+      shifted = %{ shifted | :year => shifted.year - 1, :month => 12 }
+    end
+
     # If the shift fails, it's because it's a high day number, and the month
     # shifted to does not have that many days. This will be handled by always
     # shifting to the last day of the month shifted to.
--- a/lib/datetime/datetime.ex
+++ b/lib/datetime/datetime.ex
@@ -793,9 +793,10 @@ defmodule Timex.DateTime do
       :else  -> %{datetime | :month => m}
     end
 
-    # setting months to remainders may result in month = 0
-    if shifted.month == 0 do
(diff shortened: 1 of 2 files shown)


mindreframer:
@bitwalker Hey Paul, would you mind merging this pull request? We're using this fork in production because it is well... broken. And we need to shift months reliably... Is anything missing here before you would consider merging it?

lib/datetime/datetime.ex
```diff
@@ -792,6 +792,12 @@ defmodule Timex.DateTime do
       m < 0  -> %{datetime | :year => year + div(m, 12), :month => 12 + rem(m, 12)}
       :else  -> %{datetime | :month => m}
     end
+
+    # setting months to remainders may result in month = 0
+    if shifted.month == 0 do
+      shifted = %{ shifted | :year => shifted.year - 1, :month => 12 }
```
bitwalker: This kind of imperative-style binding code doesn't work in newer versions of Elixir. Could you change it to:

```elixir
shifted = case shifted.month do
  0 -> %{ shifted | :year => shifted.year - 1, :month => 12 }
  _ -> shifted
end
```

bitwalker:
One small adjustment, and then I can merge :)

mindreframer:
@bitwalker good eye! You're right, that would not work with coming scope changes in Elixir 1.3... Let's wait for @gutschilla to fix it... :)

gutschilla:
Thanks for pointing that out. Gimme a few hours...

gutschilla:
OK, did it. Again, thanks for pointing me there.

gutschilla:
Hey Paul, anything left open in this case? Thanks!

bitwalker:
I'm going to wait until #175 is merged first, which should be very soon. I'll get this released this weekend!

bitwalker:
Alright, we're good to go :). I'll get this released this weekend.

gutschilla:
Thanks a lot for your efforts, Paul!

PR #167, opened 2016-05-03
fix ampm parsing (12am and 13am)

casellepete:
### Summary of changes

`Timex.parse("12 am", "{h12} {am}")`  now returns T00:00 instead of noon.

`Timex.parse("17 am", "{h24} {am}")` now returns T17:00 instead of T05:00.

### Checklist

- [X ] New functions have typespecs, changed functions were updated
- [X ] Same for documentation, including moduledocs
- [X ] Tests were added or updated to cover changes
- [X ] Commits were squashed into a single coherent commit

### DIFF
--- a/lib/parse/datetime/parser.ex
+++ b/lib/parse/datetime/parser.ex
@@ -226,17 +226,15 @@ defmodule Timex.Parse.DateTime.Parser do
       :us -> %{date | :millisecond => div(value, 1000)}
       :sec_epoch -> DateTime.from_seconds(value, :epoch)
       am_pm when am_pm in [:am, :AM] ->
-        {converted, hemisphere} = Time.to_12hour_clock(hh)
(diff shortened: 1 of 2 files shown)


casellepete:
I am new to coveralls, so I may be mistaken, but is coveralls reporting that coverage decreased on the file lib/time/time.ex?  Is it reporting that line 84 (def to_seconds) went from being tested one time (1x) in the previous commit, to not being tested at all in this commit?  Don't lines 84-90 of test/time_test.ex test that function multiple times?

bitwalker:
I wouldn't worry about it, coveralls is not always 100% accurate with coverage I've found. A lot of the time the difference comes from the total number of lines changing, and the uncovered portion representing a larger amount of what remains. Anyway, the code looks good to me, so I'm merging regardless.

Issue #174, opened 2016-05-19
Timex.shift/2 gives wrong year when shifting by negative months over a year boundary


nickgartmann:
But only if it is a shift less than 12 months, and doesn't end up in December (12)

### Steps to reproduce

- `Timex.shift/2`
- `%Timex.Date{years: 2016, months: 5, days: 19}, months: -6`
```
Timex.shift(%Timex.Date{years: 2016, months: 5, days: 19}, months: -6)
Expected: #<Date(2015-11-19)> 
Actual: #<Date(2016-11-19)>
```

### Description of Issue
This works for any negative month number that causes a result in November or earlier in the previous year. It works as intended for results in December (using `months: -5` gives the expected result of `#<Date(2015-12-19)>` in this case).

Some other interesting tidbits, I get the expected result when shifting by `-12` months, but `-12` gives the same result as `-24`.

bitwalker:
Fixed in #175

PR #175, opened 2016-05-20
fix shifting bug

mitchellhenke:
### Summary of changes

Fix for #174.  The test added should fail without this change.  The issue is `div(-3, 12)` (or any negative number greater than -12), will return 0 for the year shift, when it should go back a year.

### Checklist

### DIFF
--- a/lib/datetime/datetime.ex
+++ b/lib/datetime/datetime.ex
@@ -789,7 +789,7 @@ defmodule Timex.DateTime do
       value == -12 -> %{datetime | :year => year - 1}
       m == 0 -> %{datetime | :year => year - 1, :month => 12}
       m > 12 -> %{datetime | :year => year + div(m, 12), :month => rem(m, 12)}
-      m < 0  -> %{datetime | :year => year + div(m, 12), :month => 12 + rem(m, 12)}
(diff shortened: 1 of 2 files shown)


lib/datetime/datetime.ex
```diff
@@ -789,7 +789,7 @@ defmodule Timex.DateTime do
       value == -12 -> %{datetime | :year => year - 1}
       m == 0 -> %{datetime | :year => year - 1, :month => 12}
       m > 12 -> %{datetime | :year => year + div(m, 12), :month => rem(m, 12)}
-      m < 0  -> %{datetime | :year => year + div(m, 12), :month => 12 + rem(m, 12)}
+      m < 0  -> %{datetime | :year => year + Enum.min([div(m, 12), -1]), :month => 12 + rem(m, 12)}
```
bitwalker: You can change `Enum.min([div(m, 12), -1])` to `min(div(m, 12), -1)`, the latter should be much faster since it's a simple comparison of two integer values, rather than constructing a list, etc.

bitwalker:
One small fix and I'll get this merged. Thanks for the PR!

mitchellhenke:
updated!

bitwalker:
Perfect, thanks!

PR #180, opened 2016-06-01
Adding Translation to Portuguese

tubedude:
### Summary of changes

Added translation to portuguese.


bitwalker:
Thank you for the PR!

tubedude:
@bitwalker I couldn't understand the reason why the coverage decreased. I looked at other pull requests adding translation and no one added tests to it except the russian examples in the documentation. Do you recommend any action here?

bitwalker:
The only reason is that the total number of lines of code in the project increased, but there was no additional lines of code covered by tests. It's totally normal and nothing to worry about :)

PR #183, opened 2016-06-09
Fix and improve quickfast introduction

migore:
Added code highlighting to the quickfast introduction so it's better to read.
Extracted two variables using pattern matching that are used later and I got an error when running through the quickfast.


bitwalker:
Thanks for the PR!

PR #189, opened 2016-06-27
examples were not rendered fine in the documentation

edmz:
### Summary of changes

Indentation was wrong, so examples were not rendered correctly in the docs.

### Checklist

It was due to incorrect align of markdown.


bitwalker:
Fixed in master

PR #192, opened 2016-06-29
Add how to use 'after?' to README

kenta-s:
The README does not contain what I was looking for; how to use Timex.after?/2

I think it should be included as well because many people(including me) want to use it when they implement 'New Product!' things for their applications.


bitwalker:
I had to add this manually since master is so significantly diverged. Thanks for the help!

PR #197, opened 2016-07-19
Remove unneeded earmark dependency

TBK145:
From ex_doc 0.12.0, earmark is no longer an optional dependency and doesn't have to be listed separately anymore. This can be proved by the fact that, while I removed earmark from the mix.exs, it's still present in the mix.lock.


bitwalker:
Nice! Thanks!

PR #202, opened 2016-07-30
Spanish translations

marcelog:
Hello,

This adds support for the Spanish language.

There is one catch, though.

In English, May (short name) and May (long name) are the same. But in Spanish (at least), May (short name) and Mayo (long name) are different. 

The file months.po does not have a placeholder for the long name for May. I'm not really sure how to tackle this, but it seems the same thing happens in Italian, where Mag and Maggio are different but there's only one placeholder. 

Any thoughts?

Thank you!

### DIFF
--- a/priv/translations/es/LC_MESSAGES/day_periods.po
+++ b/priv/translations/es/LC_MESSAGES/day_periods.po
@@ -1,0 +1,28 @@
+## This file is a PO Template file.
+##
+## `msgid`s here are often extracted from source code.
+## Add new translations manually only if they're dynamic
+## translations that can't be statically extracted.
+##
+## Run `mix gettext.extract` to bring this file up to
+## date. Leave `msgstr`s empty as changing them here as no
+## effect: edit them in PO (`.po`) files instead.
+msgid ""
+msgstr ""
+"Language: es\n"
+
+#: lib/l10n/translator.ex:183
+msgid "AM"
+msgstr "AM"
+
+#: lib/l10n/translator.ex:185
+msgid "PM"
+msgstr "PM"
+
+#: lib/l10n/translator.ex:184
+msgid "am"
+msgstr "am"
+
+#: lib/l10n/translator.ex:186
+msgid "pm"
+msgstr "pm"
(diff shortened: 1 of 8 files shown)


bitwalker:
We should probably make sure May has both short/long variants, I can take a look at this and update once I have a bit more of an idea of how to proceed.

bitwalker:
So I just reviewed the `Translator` module, and it has the following:

```elixir
  dgettext "months", "Jan"
  dgettext "months", "Feb"
  dgettext "months", "Mar"
  dgettext "months", "Apr"
  dgettext "months", "May"
  dgettext "months", "Jun"
  dgettext "months", "Jul"
  dgettext "months", "Aug"
  dgettext "months", "Sep"
  dgettext "months", "Oct"
  dgettext "months", "Nov"
  dgettext "months", "Dec"

  dgettext "months", "January"
  dgettext "months", "February"
  dgettext "months", "March"
  dgettext "months", "April"
  dgettext "months", "May"
  dgettext "months", "June"
  dgettext "months", "July"
  dgettext "months", "August"
  dgettext "months", "September"
  dgettext "months", "October"
  dgettext "months", "November"
  dgettext "months", "December"
```

These are the "short" and "long" versions. As far as I can tell, there is support for the two variants you mentioned. Is that what you meant?

marcelog:
Yes, but this file: https://github.com/bitwalker/timex/blob/master/priv/translations/months.pot#L79 has only 1 placeholder with ID "May". So the "long may" should be changed then to something like "LongMay" and then add this new msgid to all the translations available, right?

bitwalker:
Oh ha, did not even see that. Ok yeah, I'll look into changing this, just have to make sure there's no code expecting "May" to be translated based on a parse or something.

marcelog:
Hi @bitwalker! I know this might not be a priority right now, but I was wondering if you have any plans on going back to this pull.

Thank you! :)

Best,

bitwalker:
@marcelog I've pushed a change to master which introduces a different msgid for abbreviated months, and updates existing translations to use it. Could you rebase your translations on the new .pot files and add in the translations you couldn't before? Once that's done I can go ahead and merge :)

marcelog:
Updated :) Thank you @bitwalker!

bitwalker:
No problem, but you deserve the thanks here @marcelog :)

marcelog:
My pleasure! :)

PR #211, opened 2016-08-18
Correct an error in the example

jeffrafter:
### Summary of changes

Simple README change

README has an invalid example... should be `false`


bitwalker:
Thanks!

PR #217, opened 2016-09-23
Fix relative formatter documation

iamvery:
### Summary of changes

I had some trouble figuring out how to use the relative formatter initially. First, the docs were pretty buried. Then, when I found them, they had a mistake. So I:

- added a valid example to the readme
- fixed the broken example in the formatter module
- added doctest to prove example correctness

Not sure how you'll feel about the doctest sense there is no precedence. Might be worth doing this for all the modules. I can take a stab if you're interested :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,9 @@ Here's a few simple examples:
 > {:ok, default_str} = Timex.format(datetime, "{ISO:Extended}")
 {:ok, "2016-02-29T12:30:30.120+00:00"}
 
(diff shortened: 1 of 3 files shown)


iamvery:
Also btw, thank you for all your hard work integrating with the new calendar types so quickly. It's a really big integration improvement with other libs, e.g. Ecto (rc)

bitwalker:
Looks good! Thanks for the PR! And no problem by the way, we all benefit when things play nice together :)

PR #222, opened 2016-10-03
Converting pt files to UTF-8 file encoding

douglascorrea:
### Summary of changes

It was failing when translating Portuguese values which had accents like "á" because the encoding of po files.
So I converted to utf-8 (it was iso-8859-1)


bitwalker:
Thanks! I didn't notice those were encoded wrong, good catch :)

PR #224, opened 2016-10-04
Brazilian Portuguese added

douglascorrea:
### Summary of changes

Besides for Timex the translations are the same, pt and pt_BR are different locales and usually we will use pt_BR translations on gettext, than it is easier to use Gettext.get_locale as input parameter of Timex.lformat when needed.

I also think that we should use the ll_CC standard (https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html) for all translations, but it will probably be a breaking changes, this is way I'm doing this only for Brazilian and let pt/ folder as is.

### DIFF
--- a/priv/translations/pt_BR/LC_MESSAGES/day_periods.po
+++ b/priv/translations/pt_BR/LC_MESSAGES/day_periods.po
@@ -1,0 +1,36 @@
+# # `msgid`s in this file come from POT (.pot) files.
+# #
+# # Do not add, change, or remove `msgid`s manually here as
+# # they're tied to the ones in the corresponding POT file
+# # (with the same domain).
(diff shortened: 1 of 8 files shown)


bitwalker:
Thanks for the PR! This is great! I'll review the link and think about how we could move to that format if it makes sense to do so (I'm sure it does, I just need to make sure I understand it).

douglascorrea:
@bitwalker will you release it? I'm using it as a commit ref in my mix.exs but when it is more complicated to deploy in heroku using mix file with git repo references

bitwalker:
@douglascorrea Yeah I'll try and get a release out ASAP, I have a few other things I would like to include.

PR #225, opened 2016-10-05
Fix subtract for small durations

gabrielelana:
Smaller durations (less than a second) where added instead of subtracted


bitwalker:
Thanks!

PR #226, opened 2016-10-11
To datetime with timezone

jadlr:
This changes the behavior of `to_datetime` when called with a `DateTime` struct.

I'm writing an application that needs to convert datetimes to other timezones and I didn't know about `Timex.Timezone.convert/2`

I was surprised by the current behavior:

```elixir
iex(1)> Timex.to_datetime(Timex.now, "Europe/Berlin") 
#<DateTime(2016-10-11T16:05:08.439973Z Etc/UTC)>
```

I was even more surprised by:

```elixir
iex(2)> Timex.to_datetime(Timex.now, "Hello")        
#<DateTime(2016-10-11T16:05:45.608048Z Etc/UTC)>
```

I think the more logical thing to do is to convert the date if the timezone differs from the one in the `DateTime` struct.

What do you think? Is there something I am overlooking?

### DIFF
--- a/lib/datetime/datetime.ex
+++ b/lib/datetime/datetime.ex
@@ -39,7 +39,7 @@ defimpl Timex.Protocol, for: DateTime do
   def to_date(date), do: DateTime.to_date(date)
 
   @spec to_datetime(DateTime.t, timezone :: Types.valid_timezone) :: DateTime.t | {:error, term}
-  def to_datetime(%DateTime{time_zone: time_zone} = d, timezone) when time_zone == timezone, do: d
+  def to_datetime(%DateTime{time_zone: timezone} = d, timezone), do: d
   def to_datetime(%DateTime{} = d, timezone), do: Timezone.convert(d, timezone)
 
   @spec to_naive_datetime(DateTime.t) :: NaiveDateTime.t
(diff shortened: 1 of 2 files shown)


lib/datetime/datetime.ex
```diff
@@ -39,7 +39,8 @@ defimpl Timex.Protocol, for: DateTime do
   def to_date(date), do: DateTime.to_date(date)
 
   @spec to_datetime(DateTime.t, timezone :: Types.valid_timezone) :: DateTime.t | {:error, term}
-  def to_datetime(%DateTime{} = d, _timezone), do: d
+  def to_datetime(%DateTime{time_zone: time_zone} = d, timezone) when time_zone == timezone, do: d
```
bitwalker: You can just do `to_datetime(%DateTime{time_zone: timezone} = d, timezone), do: d` which will only match if both the provided timezone and the one on the DateTime are the same.

bitwalker:
You're definitely correct, your changes look good except for the one tweak I suggested. I'll get this merged once that's updated. Thanks for the PR!

jadlr:
@bitwalker thanks for the hint! I didn't know that. 

I just found out that `Timex.Timezone.convert` can return an `AmbiguousDateTime`. Could that be a Problem? Should I add `AmbiguousDateTime` to the `@spec` of `to_datetime`?

bitwalker:
Yeah could you add it to the `@spec`, it should be there. AmbiguousDateTime is intended to be resolved by the caller, but it should only occur with DateTime, no other types, so we need to pass it back to the caller for any conversions like this which may result in an AmbiguousDateTime.

bitwalker:
Thanks again!

PR #229, opened 2016-10-25
bump gettext version

### LINKED ISSUE #228: gettext 0.12.x breaks Timex 3.x
### Steps to reproduce

- API function called: `Timex.format!()`
- Arguments provided: `"{Mfull} {D} {YYYY}, {h12}:{m}{am}"`

### Description of issue

Timex is calling a private gettext API, which changed in the past two days with their 0.12.0 release.

### Stacktrace

(I'm on Timex 3.1.0)

```
** (exit) an exception was raised:
    ** (UndefinedFunctionError) function Timex.Gettext.lgettext/3 is undefined or private. Did you mean one of:

      * lgettext/4

        (timex) Timex.Gettext.lgettext("en", "months", "January")
        (timex) lib/l10n/translator.ex:156: Timex.Translator.get_domain_text/3
        (timex) lib/l10n/translator.ex:103: Timex.Translator.get_months/1
        (timex) lib/format/datetime/formatter.ex:441: Timex.Format.DateTime.Formatter.format_token/6
        (timex) lib/format/datetime/formatters/default.ex:182: Timex.Format.DateTime.Formatters.Default.do_format/4
        (timex) lib/format/datetime/formatter.ex:54: Timex.Format.DateTime.Formatter.lformat!/4
```

### Temp Resolution

Forcing my mix.deps to use the `0.11.0` version made this error go away.

jadlr:
### Summary of changes

use `lgettext/4` instead of `lgettext/3`, since no bindings are needed an empty map does the trick.

fixes #228 

### Checklist

### DIFF
--- a/lib/l10n/translator.ex
+++ b/lib/l10n/translator.ex
@@ -153,7 +153,7 @@ defmodule Timex.Translator do
 
   @spec get_domain_text(locale :: String.t, domain :: String.t, msgid :: String.t) :: String.t
   defp get_domain_text(locale, domain, msgid) do
-    case Timex.Gettext.lgettext(locale, domain, msgid) do
+    case Timex.Gettext.lgettext(locale, domain, msgid, %{}) do
       {:ok, translated}   -> translated
       {:default, default} -> default
     end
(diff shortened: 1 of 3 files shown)


bitwalker:
Thanks!

james-vanlommel-zefr:
JFYI, when I first mentioned this issue on #elixir-lang, Jose asked why timex was using an internal API of gettext - if there's a continued need for `lgettext`, it may worth opening up an issue or PR on gettext so future internal changes don't cause breaks in Timex.

bitwalker:
I'm not sure how it's private, since it's exposed via the gettext module generated via the `__using__` macro, but I'll review it to see if maybe we're using it wrong.

jadlr:
@bitwalker I think the answer lies in the `locale` paragraph of the docs: https://hexdocs.pm/gettext/Gettext.html#module-locale

TL;DR: You're supposed to store the locale per process via `Gettext.put_locale/2`

bitwalker:
Interesting that they don't support explicitly passing the locale in the public API. I liked the explicit route since it was clear what was being used. Looks like the fix is to wrap formatting calls in something which pops the current locale, pushes the provided one, then pushes the previous locale back. The API could support the implicit locale as well (iirc, it does that anyway, but I'd need to double check).

PR #232, opened 2016-10-27
Fix typo in typespec for Timex.shift

obrok:
### Summary of changes

The typespec for `Timex.shift` used `Types.valid_datetimes` but the type defined there is `Types.valid_datetime`. This caused dialyzer to generate a warning in downstream projects.

Is there some reason why dialyzer is not part of the travis build? I could try to assemble a PR  adding that if you're interested.

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -1413,7 +1413,7 @@ defmodule Timex do
     weeks: integer,
     years: integer
   ]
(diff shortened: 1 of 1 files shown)


bitwalker:
There is no specific reason, other than I haven't yet spent time to iron out dialyzer warnings entirely.

PR #234, opened 2016-10-28
Fix shifting into December

obrok:
### Summary of changes

When shifting into December the computation would set the month to 0 which in turn caused a crash:

```elixir
** (FunctionClauseError) no function clause matching in :calendar.last_day_of_the_month1/2
     stacktrace:
       (stdlib) calendar.erl:244: :calendar.last_day_of_the_month1(2001, 0)
       (timex) lib/datetime/naivedatetime.ex:248: Timex.Protocol.NaiveDateTime.shift_by/3
       (timex) lib/datetime/naivedatetime.ex:213: Timex.Protocol.NaiveDateTime.apply_shifts/2
       (timex) lib/date/date.ex:189: Timex.Protocol.Date.shift/2
       test/shift_test.exs:79: (test)
```


bitwalker:
Thanks!

PR #235, opened 2016-10-31
Add months to typespec for Timex.shift

obrok:
### Summary of changes

The option list typespec didn't allow months despite it clearly working. Sorry for not noticing this when I changed this earlier.


bitwalker:
Thanks!

Issue #236, opened 2016-11-01
Timex 2.2.1 fails to format RFC1123 dates


mspanc:
### Steps to reproduce

Under timex 2.2.1, which is the last supported on Elixir 1.2, the following fails:

(record is some Ecto model)

- record.updated_at |> Ecto.DateTime.to_iso8601 |> Timex.parse!("{ISO:Extended}") |> Timex.format!("{RFC1123}")

Causes UndefinedFunctionError because it internally calls

- Timex.Gettext.lgettext
- `["en", "months", "Jan"]`

but gettext 0.12.1 that is used in 2.2.1 requires 4 arguments to lgettext.

bitwalker:
If you lock the version of gettext at 0.11, this issue won't affect you with 2.2.1. 3.x already has this fixed, and it affects 2.2.1 because it's version restriction on gettext is very loose.

PR #237, opened 2016-11-01
Fix for https://github.com/bitwalker/timex/issues/236

mspanc:
Fix for https://github.com/bitwalker/timex/issues/236


bitwalker:
This change is already present in master, and has been in the last few 3.x releases.

mspanc:
Don't you think that releasing fix for apps that are stuck with elixir 1.2
would make sense? Not everyone is ready to follow the master branch just
for sake of following the master.

bitwalker:
It's been on hex in the 3.1.x series of releases as well, I'm not suggesting tracking against master. That said, since gettext is the breaking change here, it's better to lock in to gettext 0.11 until you can upgrade to elixir 1.3+ and thus timex 3.x+.

PR #239, opened 2016-11-04
fix typespec

sasa1977:


sasa1977:
Not sure why the coverage decreased? I didn't add any line of code.

bitwalker:
No worries, must be a problem with how coverage is calculated

PR #245, opened 2016-11-17
fix @doc formatting

techgaun:
### Summary of changes

Changed spacing so the doc formatting looks good in `mix docs` generation.

### Checklist


techgaun:
Also, I was wondering the reason for using `__MODULE__` in many places. There are inconsistencies in the choice. I thought to change those as well but wanted to make sure that there's a reason behind it

bitwalker:
I mainly use the macro in places where the module name is long, but there are still legacy instances of it around in places where I wouldn't necessarily use it today. Thanks for the PR!

PR #247, opened 2016-11-17
Enables fallback_to_any in Timex.Protocol

### LINKED ISSUE #246: Timex.Protocol does not fallback to Any
### Steps to reproduce

```
$iex -S mix
# This works
iex()> Timex.is_valid?({2016, 01, 01})
true

# This should be {:error, :invalid_date}
iex()> Timex.is_valid?("caturday")
** (Protocol.UndefinedError) protocol Timex.Protocol not implemented for "caturday"
    (timex) lib/protocol.ex:1: Timex.Protocol.impl_for!/1
    (timex) lib/protocol.ex:150: Timex.Protocol.is_valid?/1
```

### Description of issue

- What are the expected results?
   The call to Timex.is_valid?("caturday") should return either false or the tuple {:error, :invalid_date}. 
   Definitely not throwing an exception. Also from the code in https://github.com/bitwalker/timex/blob/master/lib/protocol.ex#L163 it seems the intention is to fallback to returning an :error tuple. This happens also with other functions defined in the protocol

- If you are certain this is a bug in Timex, do you have references I
  can use to check the algorithm/logic against? 
  
 In the definition of Timex.Protocol should be added the line 

    @fallback_to_any true

according do the [elixir docs](http://elixir-lang.org/getting-started/protocols.html#fallback-to-any)

voger:
### Summary of changes
fixes #246

Unfortunately I don't have much experience with git so I couldn't squash the commits properly. I hope I will do better next time. 

There are two commits. One for code and one for tests. I hope this is not a big problem.

### Checklist

- [ NA] New functions have typespecs, changed functions were updated
- [ NA] Same for documentation, including moduledocs
- [ Y] Tests were added or updated to cover changes
- [ N] Commits were squashed into a single coherent commit

### DIFF
--- a/lib/protocol.ex
+++ b/lib/protocol.ex
@@ -4,6 +4,8 @@ defprotocol Timex.Protocol do
   `NaiveDateTime`, or `DateTime` as input.
   """
 
+  @fallback_to_any true
+
   @doc """
   Convert a date/time value to a Julian calendar date number
   """
--- a/test/timex_test.exs
+++ b/test/timex_test.exs
@@ -43,6 +43,8 @@ defmodule TimexTests do
     localdate = {{2013,3,17},{11,59,10}}
     assert Timex.weekday(Timex.to_datetime(localdate)) === 7
     assert Timex.weekday(Timex.epoch()) === 4
+    assert {:error, :invalid_date} = Timex.weekday("Made up date")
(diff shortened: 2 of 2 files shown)


bitwalker:
Thanks for the PR!

PR #257, opened 2016-12-29
Fix from_iso_triplet logic

mbramson:
Shoutout to https://github.com/ericworkman who paired with me on all of this.

When implementing a feature that used `from_iso_triplet` to get the first day of an iso week, a case was found around the boundary of the year where `from_iso_triplet` generated a match error when it called `iso_day_to_date_tuple`. This is because before `from_iso_triplet` calls this method it calculates an offset for the day, which it then adds to the day, which can result in a day that is < 1 or greater than > 366. `iso_day_to_date_tuple`'s guard clause prohibits such days.

The fix to that problem was to make a conversion from ordinal_day after the offset has been applied to iso day, accounting for all situations where the ordinal_day crosses the year boundary (taking into account leap years).

In the process of implementing this fix, a bug was found in the implementation of `iso_day_to_date_tuple` that resulted in wrong resulting dates. Tests around all boundaries were added and ultimately, it was determined that the guard clause should be tighter to account for leap and non-leap years. So a day of 366 is no longer a valid input to `iso_day_to_date_tuple` if the year is a non-leap year, but it is if the year is a leap year.

Guard macros for `is_leap_year` and `is_iso_day_of_year` were added to deal with this, and ultimately should be good utility functions for the public Timex API. 

from_iso_triplet now works for dates that cross year boundaries.

The updated `iso_day_to_date_tuple` method now errors with error tuples
for invalid days (< 1 or > 366, depending on leap years) rather than
match errors, so the input to that function needs to be controlled more
tightly. An ordinal date to year and iso_day conversion step was added to
ensure this. When the day or year is invalid for `iso_day_to_date_tuple`, an appropriate error tuple is now returned rather than simple a match error that is less than informative. I tried to match the surrounding style of error tuples, but please let me know if you think they should be more specific (perhaps indicating by which direction the day or year is invalid).

Of note, I think that the `iso_day_to_date_tuple` function might not belong in Helpers any longer, as it could be a useful function for the public Timex API, and its certainly robust enough now.  That said, I think that's a decision for a separate PR.

Single Commit:

Add test cases around iso_day_to_date_tuple

Add is_leap_year guard macro

Add is_iso_day_of_year guard macro

Change iso_day_to_date_tuple guard clause to use is_iso_day_of_year
instead of is_day_of_year to more correctly bound possible matched input
values. This allows the removal of the first cond in
iso_day_to_date_tuple.

Add error tuples to iso_day_to_date_tuple for days and years which are
invalid.

### Summary of changes

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -620,8 +620,19 @@ defmodule Timex do
       do
       {_, _, jan4weekday} = iso_triplet({year, 1, 4})
       offset = jan4weekday + 3
-      ordinal_date = ((week * 7) + weekday) - offset
-      {year, month, day} = Helpers.iso_day_to_date_tuple(year, ordinal_date)
+      ordinal_day = ((week * 7) + weekday) - offset
+      {year, iso_day} = case {year, ordinal_day} do
+        {year, ordinal_day} when ordinal_day < 1 and is_leap_year(year - 1) ->
+          {year - 1, ordinal_day + 366}
+        {year, ordinal_day} when ordinal_day < 1 ->
+          {year - 1, ordinal_day + 365}
+        {year, ordinal_day} when ordinal_day > 366 and is_leap_year(year) ->
+          {year + 1, ordinal_day - 366}
+        {year, ordinal_day} when ordinal_day > 365 and not is_leap_year(year) ->
+          {year + 1, ordinal_day - 365}
+        _ -> {year, ordinal_day}
+      end
+      {year, month, day} = Helpers.iso_day_to_date_tuple(year, iso_day)
       %Date{year: year, month: month, day: day}
   end
   def from_iso_triplet(_, _, _), do: {:error, {:from_iso_triplet, :invalid_triplet}}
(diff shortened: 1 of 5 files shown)


mbramson:
It appears that CI is failing because coverage decrease by two lines in lib/time/duration.ex. As this PR did not touch that file, my conclusion is that some of the logic that was removed resulted in tests that existed previously (and still do, I didn't remove any tests) no longer hitting something in that file.  A potential candidate is my not using the `is_day_of_year` guard macro for `iso_day_to_date_tuple` in favor of `is_iso_day_of_year`. I haven't isolated why exactly, but as code and case coverage improved in the parts of the code that I did touch, I'd think that this is an acceptable tradeoff. What are your thoughts @bitwalker ?

bitwalker:
This is excellent work! Thanks for the explanation and detail, it helps me review things a lot faster :)

As for the coverage, it's because the total line count increased, but the number of covered lines did not increase proportionally. It's nothing to worry about, I've been meaning to fix that for awhile.

PR #258, opened 2016-12-30
Add doctests for Timex.Duration

mbramson:
No active code change.

Performed a documentation and doctest pass on the Timex.Duration module. Added doctests to any public function that did not have test coverage in test/duration_test.exs (or in existing doctests. Tweaked any documentation that appeared inaccurate.

Single Commit:

Add doctests to all functions in Timex.Duration that do not have any
associated unit tests.

Alter a few documentation strings that were inaccurate

### Summary of changes

### Checklist

### DIFF
--- a/lib/time/duration.ex
+++ b/lib/time/duration.ex
@@ -26,6 +26,12 @@ defmodule Timex.Duration do
 
   @doc """
   Converts a Duration to an Erlang timestamp
+
+  ## Example
+
+      iex> d = %Timex.Duration{megaseconds: 1, seconds: 2, microseconds: 3}
+      ...> Timex.Duration.to_erl(d)
+      {1, 2, 3}
   """
   @spec to_erl(__MODULE__.t) :: Types.timestamp
   def to_erl(%__MODULE__{} = d),
@@ -33,14 +39,29 @@ defmodule Timex.Duration do
 
   @doc """
   Converts an Erlang timestamp to a Duration
+
+  ## Example
+
+      iex> Timex.Duration.from_erl({1, 2, 3})
+      %Timex.Duration{megaseconds: 1, seconds: 2, microseconds: 3}
   """
   @spec from_erl(Types.timestamp) :: __MODULE__.t
   def from_erl({mega, sec, micro}),
     do: %__MODULE__{megaseconds: mega, seconds: sec, microseconds: micro}
 
   @doc """
-  Converts a Duration to a Time, if the duration fits within a 24-hour clock,
-  if it does not, an error will be returned.
+  Converts a Duration to a Time if the duration fits within a 24-hour clock.
+  If it does not, an error tuple is returned.
+
+  ## Examples
+
+      iex> d = %Timex.Duration{megaseconds: 0, seconds: 4000, microseconds: 0}
+      ...> Timex.Duration.to_time(d)
+      {:ok, ~T[01:06:40]}
+
+      iex> d = %Timex.Duration{megaseconds: 1, seconds: 0, microseconds: 0}
+      ...> Timex.Duration.to_time(d)
+      {:error, :invalid_time}
   """
   @spec to_time(__MODULE__.t) :: {:ok, Time.t} | {:error, atom}
   def to_time(%__MODULE__{} = d) do
@@ -49,7 +70,18 @@ defmodule Timex.Duration do
   end
 
   @doc """
-  Same as to_time/1, but returns the Time directly, and raises on error
+  Same as to_time/1, but returns the Time directly. Raises an error if the
+  duration does not fit within a 24-hour clock.
+
+  ## Examples
+
+      iex> d = %Timex.Duration{megaseconds: 0, seconds: 4000, microseconds: 0}
+      ...> Timex.Duration.to_time!(d)
+      ~T[01:06:40]
+
+      iex> d = %Timex.Duration{megaseconds: 1, seconds: 0, microseconds: 0}
+      ...> Timex.Duration.to_time!(d)
+      ** (ArgumentError) cannot convert {277, 46, 40} to time, reason: :invalid_time
   """
   @spec to_time!(__MODULE__.t) :: Time.t | no_return
   def to_time!(%__MODULE__{} = d) do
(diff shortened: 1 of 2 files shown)


lib/time/duration.ex
```diff
@@ -424,12 +545,25 @@ defmodule Timex.Duration do
...
+  - :hours
+  - :days
+  - :weeks
+
+  ## Examples
 
-  The argument is an atom indicating the type of time units to return (see
-  convert/2 for supported values).
+  iex> Timex.Duration.now(:seconds)
+  :os.system_time(:seconds)
```
mbramson: So this is a terrible test is it stands as it's purely tautological. That said, it at least demonstrates the behavior for documentation purposes, so I left it in.

lib/time/duration.ex
```diff
@@ -424,12 +545,25 @@ defmodule Timex.Duration do
...
 
-  When the argument is omitted, the return value's format is { megasecs, seconds, microsecs }.
+  When the argument is omitted or nil, a Duration is returned.
```
mbramson: I would love to have a doctest after this, but I don't know of a way to write an elegant one.
```
iex> Timex.Duration.now()
:os_system_time(:micro_seconds) |> from_microseconds
```

would be purely tautological as well, and does a worse job than the above tautological test at explaining how this works. I don't like it. Any thoughts on a better way to do this?

mbramson:
This was an interesting PR. A few thoughts that came out of this:
- There are a number of functions that return negative durations.
  - Does a negative duration make sense? I wouldn't want to throw away negative durations because they are probably quite useful as intermediate steps in larger calculations, but does that mean a negative duration is justified as an end result?
  - `Timex.Format.Duration.Formatter.format(duration)` errors for some negative durations (interestingly, not all, but the output is messy when an error doesn't occur). This should probably be addressed in a separate issue/PR. Could be a nice beginner level issue for someone to munch on.
- `Timex.Duration.normalize/1` as stated in the TODO comment needs to be looked at for negative values. I didn't poke it too heavily, but it looks like it could us some looking at for various permutations of negative and large values of the Duration key-values. Probably another fairly beginner issue for someone to pick up.
- A number of functions let you specify time units, but stop at :weeks. What about :years? (:months might be a bit wonky). Another good issue for someone.
- abs and invert (and possibly others) could use some analysis around when they receive input durations where the various key-values are not all-positive or all-negative. Probably a more intermediate level issue there depending on what is found.

lib/time/duration.ex
bitwalker: I think this test is prone to a race condition, which may be rare enough to not be a problem, but philosophically I don't like it. I feel like a better test is whether or not the value returned makes sense for the given date, e.g. `:os.system_time(:seconds) <= :calendar.datetime_to_gregorian_seconds({{1970,1,1},{0,0,0}}) + Timex.Duration.now(:seconds)`

lib/time/duration.ex
bitwalker: Same as above, I think the best test is one which verifies that the value returned is valid for the current date/time, i.e. it's not above or below a range which is reasonable for covering the inherent race condition present in the function.

bitwalker:
Could you create an issue for that formatter error you found? Definitely a bug. Negative durations are considered valid, since one place where durations can be generated is relative to another point in time, thus depending on whether it's past/future, the value may be negative or positive.

I'll create an issue for looking at `normalize`, and tag it for help wanted, definitely a good place for someone looking to get familiar with the code.

Measurements in months or years have to make a decision about what the length of a month or a year is, and that's not set in stone (i.e. years can vary based on whether they are leap years or not), months vary depending on the number of days in a given month, where the length of a week is static (i.e. 7 days). One can assign reasonable static values (i.e. 365 days to a year, 30 days to a month), but it's misleading, and thus I didn't want to take that step. One can easily convert a number of days to either depending on their own criteria, so it felt like something best left up to the consumer.

I agree about `epoch`/`now`, I'll take a look and see what makes sense for them to return.

I left some comments about the tests, but once those are addressed I'll go ahead and merge. Thanks for all the hard work!

lib/time/duration.ex
mbramson: Excellent call. I'll take another look at these two tests.

mbramson:
Sounds good. I'll make some changes to those tests and create an issue for the formatter error. I'll mention you when this is updated.

mbramson:
So I modified the tests for the `now` function. As they became a bit more complicated, I didn't think that the documentation was the best place for them, so I added them to `timex/duration_test.exs`. I did think that some examples in the documentation were valuable, but obviously they would fail as doctests, so I exempted the `now` function from running doctests. @bitwalker what do you think?

mbramson:
Added an issue around the formatting bugs that were found: https://github.com/bitwalker/timex/issues/259.

test/duration_test.exs
```diff
@@ -111,5 +122,23 @@ defmodule DurationTests do
...
   end
 
+  describe "now" do
+    test "now/0 returns the duration since epoch" do
+      erlang_now_microseconds = div(:os.system_time, 1000)
+      timex_now_microseconds = Timex.Duration.now() |> Timex.Duration.to_microseconds
+      difference_microseconds = timex_now_microseconds - erlang_now_microseconds
+      assert difference_microseconds < 1000000
+      assert difference_microseconds >= 0
+    end
```
mbramson: So I still don't love these tests, but I think that this is a better test than what I had as a doctest previously. Shouldn't be vulnerable to a race condition, but it could be vulnerable to something running exceptionally slow. The maximum microsecond_difference that I observed was 32, but most came in below 10. I would be extremely surprised if it ever came over 1,000,000 microseconds without something being actually wrong.

mbramson: Do you think this test does a good enough job of asserting that a Duration struct is returned by `now/0`?

test/duration_test.exs
```diff
@@ -1,10 +1,21 @@
 defmodule DurationTests do
   use ExUnit.Case, async: true
   use Timex
-  doctest Timex.Duration
+  doctest Timex.Duration, except: [{:now, 0}, {:now, 1}]
```
mbramson: This is the only way I could find to get an example in the documentation for `now/0` and `now/1` that elegantly explained the behavior of the function. A doctest that actually runs would be clumsy, and not actually serve well as documentation, in my opinion. See the new tests in the `now` describe block of `test/duration_test.exs`. I'd love to be proven wrong on this point through a more elegant doctest that isn't subject to a race condition or tautological, however!

bitwalker: 👍

bitwalker:
I like having the tests in a proper module and keeping the example simple in the docs, there are definitely places (such as this one) where trying to kill two birds with one stone produces poorer documentation, so good call. Everything looks good to me, if you are ready for me to merge, just give me the thumbs up and I'll do so :)

Great work!

mbramson:
Go for it!

Issue #259, opened 2016-12-31
Fix formatting of negative and overflowing durations


mbramson:
### Steps to reproduce

Formatting of durations with negative components error or output nonsensical values. Some examples:

```
iex(1)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: -1, seconds: 0, microseconds: 0})
"PT-1000000S"
iex(2)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: -1, seconds: -1, microseconds: 0})
"PT-1000001S"
iex(3)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: -1, seconds: -1, microseconds: -1})
** (FunctionClauseError) no function clause matching in Timex.Format.Duration.Formatters.Default.get_fractional_seconds/3
    (timex) lib/format/duration/formatters/default.ex:93: Timex.Format.Duration.Formatters.Default.get_fractional_seconds(-1000001, -1, [])
    (timex) lib/format/duration/formatters/default.ex:51: Timex.Format.Duration.Formatters.Default.lformat/2
iex(4)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: 1, seconds: 1, microseconds: -1})
** (FunctionClauseError) no function clause matching in Timex.Format.Duration.Formatters.Default.get_fractional_seconds/3
    (timex) lib/format/duration/formatters/default.ex:93: Timex.Format.Duration.Formatters.Default.get_fractional_seconds(40, -1, [minutes: 46, hours: 13, days: 11])
    (timex) lib/format/duration/formatters/default.ex:51: Timex.Format.Duration.Formatters.Default.lformat/2
```

There is also at least one other buggy case in this function not involving negative numbers:
```
iex(1)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: 0, seconds: 1, microseconds: 1000000})
"PT3.0S"
```
(I'd expect `"PT2.0S"` or `"PT2S"`)
**NOTE: The above example when microseconds exceeds 1000000 has been fixed by https://github.com/bitwalker/timex/pull/260**

### Description of issue

I am not entirely sure what the expected output would be in these cases, but it definitely shouldn't be throwing an error. @bitwalker might be able to provide more guidance on what the expected output looks like.

bitwalker:
[...]

PR #260, opened 2017-01-01
Fix Duration formatting when microseconds amount exceeds second interval

slavakisel:
### Summary of changes

Fixes a bug with duration formatting from #259 when microseconds amount exceeds 1_000_000.

Before:
```
iex(1)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: 0, seconds: 1, microseconds: 1_000_000})
"PT3.0S"
```

After:
```
iex(1)> Timex.Format.Duration.Formatter.format(%Timex.Duration{megaseconds: 0, seconds: 1, microseconds: 1_000_000})
"PT2S"
```

### Checklist

### DIFF
--- a/lib/format/duration/formatters/default.ex
+++ b/lib/format/duration/formatters/default.ex
@@ -34,12 +34,20 @@ defmodule Timex.Format.Duration.Formatters.Default do
   ## Examples
 
       iex> use Timex
-      ...> Duration.from_erl({1435, 180354, 590264}) |> #{__MODULE__}.format
-      "P45Y6M5DT21H12M34.590264S"
+      ...> Duration.from_erl({0, 1, 1_000_000}) |> #{__MODULE__}.format
+      "PT2S"
+
+      iex> use Timex
+      ...> Duration.from_erl({0, 1, 1_000_100}) |> #{__MODULE__}.format
+      "PT2.0001S"
 
       iex> use Timex
       ...> Duration.from_erl({0, 65, 0}) |> #{__MODULE__}.format
       "PT1M5S"
+
+      iex> use Timex
+      ...> Duration.from_erl({1435, 180354, 590264}) |> #{__MODULE__}.format
+      "P45Y6M5DT21H12M34.590264S"
 
   """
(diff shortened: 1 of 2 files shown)


lib/format/duration/formatters/default.ex
```diff
@@ -74,7 +78,7 @@ defmodule Timex.Format.Duration.Formatters.Default do
   defp format_component({:seconds, s}, str), do: str <> "#{s}S"
 
   defp deconstruct(%Duration{microseconds: micro} = duration),
-    do: deconstruct({Duration.to_seconds(duration, truncate: true), micro}, [])
+    do: deconstruct({Duration.to_seconds(duration, truncate: true), rem(micro, 1_000_000)}, [])
```
slavakisel: Bug happens because `Duration.to_seconds` already includes microseconds in calculation so when microseconds amount exceeds 1 second the quotient part of microseconds is counted twice.

lib/format/duration/formatters/humanized.ex
```diff
@@ -19,12 +19,18 @@ defmodule Timex.Format.Duration.Formatters.Humanized do
...
-      iex> Duration.from_erl({0, 65, 0}) |> #{__MODULE__}.format
+
+      iex> use Timex
+      ...> Duration.from_erl({0, 65, 0}) |> #{__MODULE__}.format
       "1 minute, 5 seconds"
 
+      iex> use Timex
+      ...> Duration.from_erl({0, 1, 1_000_000}) |> #{__MODULE__}.format
+      "2 seconds"
+
```
mbramson: I might suggest reversing the order of these three doctests, so that the simplest case comes first. I think that it would make the documentation a little more educational? Does that make sense?

bitwalker: I agree, @slavakisel can you update the PR with that change?

slavakisel: @mbramson @bitwalker Sure! Please take a look at updated PR

bitwalker:
Thanks!

PR #264, opened 2017-01-11
Add i18n id support

tslcjames:
### Summary of changes

I need Indonesian translation for using timex
### Checklist


bitwalker:
Thanks!

PR #267, opened 2017-01-15
correct Russian translations

bitgamma:
### Summary of changes

The Russian plurals have been corrected for both units of measures and relative times. Also, in its full form April was in the accusative case instead of genitive like all other months. The associated documentation/tests have been corrected accordingly.

### Checklist

- [*] New functions have typespecs, changed functions were updated
- [*] Same for documentation, including moduledocs
- [*] Tests were added or updated to cover changes
- [*] Commits were squashed into a single coherent commit


bitwalker:
Thanks!

Issue #271, opened 2017-01-29
Corner case when shifting +1 months Jan 29 


marcelog:
Hello,

This is a corner case, and I'm not sure there's a workaround for it, but I'd still like to know your thoughts about it. 

Today is Jan 29th 2017. This year, February has 28 days. When adding +1 months to Jan 29, the result is Feb 28, but `Timex.diff` reports 0 months of difference.

So when doing this:

```elixir
iex(1)> {:ok, d1} = Timex.parse "Sun, 29 Jan 2017 Etc/UTC", "%a, %d %b %Y %Z", :strftime
{:ok, #<DateTime(2017-01-29T00:00:00Z Etc/UTC)>}
iex(2)> d2 = Timex.shift d1, months: 1
#<DateTime(2017-02-28T00:00:00Z Etc/UTC)>
iex(7)> Timex.diff d1, d2, :months
0
```

The result seems to be inconsistent. We're adding +1 months through the API but later on we have a diff of 0 months.

What are your thoughts?

Thanks in advance.

Best regards,

bitwalker:
Indeed this did seem to be wrong - unfortunately I did a poor job of documenting the old diffing algorithm for months, so it's not clear to me why it was the way it was, but I've resorted to a simpler algorithm which will do a naive integer division of the diff in days by 30 (we have to make an assumption about the length of a month, and 30 is a reasonable average). I believe the old algorithm attempted to be more accurate, but clearly had a flaw handling that edge case, so I've thrown it out in favor of the simpler implementation.

marcelog:
Thank you very much @bitwalker 👍

PR #281, opened 2017-02-20
Improved implementation of calculating diff in months and years

dovadi:
The fix in calculating the diff in months as a reply to the corner case mentioned in issue #271 introduces a new "bug" in calculating the diff in years, namely a year is not the same as 12 months of 30 days.

So for example the diff of {2017, 4, 10} and {2016, 4, 15} is five days less than a year and should therefor not result in 1.

### DIFF
--- a/lib/comparable/diff.ex
+++ b/lib/comparable/diff.ex
@@ -48,15 +48,37 @@ defmodule Timex.Comparable.Diff do
     end
   end
   defp do_diff(a, b, :months) do
-    div(do_diff(a, b, :days), 30)
+    {high_y, high_m, high_d, low_y, low_m, low_d, sign} = convert_to_signed_date_tuples(a, b)
+
+    nof_months = (high_y * 12 + high_m) - (low_y * 12 + low_m)
+    nof_months = if (nof_months == 0 || low_d <= high_d),  do: nof_months, else: nof_months - 1
+    nof_months * sign
   end
   defp do_diff(a, b, :years) do
-    div(do_diff(a, b, :months), 12)
+    {high_y, high_m, high_d, low_y, low_m, low_d, sign} = convert_to_signed_date_tuples(a, b)
+
+    nof_years = high_y - low_y
+    nof_years = if (nof_years == 0 || {low_y, low_m, low_d} <= {high_y - nof_years, high_m, high_d}),  do: nof_years, else: nof_years - 1
+    nof_years * sign
   end
+
   defp do_diff(_, _, granularity) when not granularity in @units,
     do: {:error, {:invalid_granularity, granularity}}
 
   defp zero(:duration), do: Duration.zero
   defp zero(_type), do: 0
+
+  defp convert_to_signed_date_tuples(a, b) do
+    cond do
+      a > b ->
+        {{high_y, high_m, high_d}, _} = :calendar.gregorian_seconds_to_datetime(div(a, 1_000*1_000))
+        {{low_y, low_m, low_d}, _}    = :calendar.gregorian_seconds_to_datetime(div(b, 1_000*1_000))
+        {high_y, high_m, high_d, low_y, low_m, low_d, 1}
+      :else ->
+        {{low_y, low_m, low_d}, _}    = :calendar.gregorian_seconds_to_datetime(div(a, 1_000*1_000))
+        {{high_y, high_m, high_d}, _} = :calendar.gregorian_seconds_to_datetime(div(b, 1_000*1_000))
+        {high_y, high_m, high_d, low_y, low_m, low_d, -1}
+    end
+  end
 
 end
--- a/test/comparable_test.exs
+++ b/test/comparable_test.exs
@@ -36,10 +36,13 @@ defmodule ComparableTests do
   end
 
   test "calculate difference in months" do
-    assert   0 == Comparable.diff({2017,4,30}, {2017,3,31}, :months)
-    assert   1 == Comparable.diff({2017,4,30}, {2017,3,30}, :months)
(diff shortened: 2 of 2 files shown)


bitwalker:
Could you add a test for the edge case mentioned in #271? I'd also like some tests around the beginning/end of the year to make sure calculations there work properly (namely diffing across the new year). Historically the bulk of the bugs in diffing/shifting have been in that region of time, so I'd like to have more coverage there for the future.

Thanks for the PR!

dovadi:
Mmm, the question is of course whether the edge case is a valid case for the diff function. At least if you compare {2017, 1,  29} and {2017, 2, 28} the difference is in my opinion not one month! Although shifting {2017,1, 29} with one month will result in {2017, 2, 28}.

bitwalker:
The difference in the former is technically 1 month, since it's 30 days between those two dates. However the problem with this has always been the fact that shifting by a month is ambiguous by definition. Timex has historically tried to do the "intuitive" thing here, but I've been slowly migrating towards whatever is "technically" correct. So, diffing by a month should always return a value of 1 if the number of days difference is greater than or equal to 30 days. Shifting by months is still exhibiting some of the older behaviour, namely shifting by months for dates on the 31st will result in a shift of 30 rather than 31 days when shifting to a month which only has 30 days, except in the case of February, where 1/30-1/31 will be shifted to 2/28 in non-leap years and will result in a shift of less than 30 days. I've never held that there should be a duality between shift and diff, but since it has come up a few times, I'm leaning towards simplifying shift to treat all months as 30 days long, such that diff will always return 1 for a shift of 1 month.

dovadi:
I agree it is also a matter of interpretation. So if you keep 'your' simplification by treating all months as 30 days long, we're still left with defining a year. With the current definition this is defined as 360 days, which doesn't seem correct.

Maybe only accepting/merging "my" implementation of the diff function for years?

bitwalker:
The definition of a year is very clear cut - it's 365 days except for leap years. The current implementation is leveraging months, which is leftovers from when Timex was trying to be smarter about diffing by months, which used to take into account variable day months as well as leap years, and we should address that (which your fix does). That said, if your new implementation matches how shift behaves (when diffing by months), then I'm ok with merging that to bring those two functions to parity with each other.

kabturek:
IMO a library should be consistent - if you shift by 3 months then then the diff should show 3 months. Simplifying a month to 30 days does this:

```
iex(13)> date = ~D[2017-04-30]
~D[2017-04-30]
iex(14)> Timex.diff(date, Timex.shift(date, months: -3), :months)
3
iex(15)> date = ~D[2017-05-15]
~D[2017-05-15]
iex(16)> Timex.diff(date, Timex.shift(date, months: -3), :months)
2 
```

So if in shift you're just changing the month then diff should show months substracted  - in this example 5 - 2 = 3.

At least you should mention this in docs for the diff method(for ex.):
- months (number of 30 day intervals)

bitwalker:
After reviewing this from start to finish, I realize that I wasn't paying close enough attention previously, and I think there is some confusion here that needs to be ironed out:

- The goal of Timex is to be accurate as possible with respect to how time is measured. As such, a year is not "12 months of 30 days", it is whatever the definition of that specific year happens to be, i.e. the number of days in that year, which is variable depending on whether it's a leap year. Likewise with months, days, etc. This will inevitably lead to some unintuitive results when diffing dates, e.g. you may think that two days are a year apart, when in fact they are not. Nevertheless, the point is that it is accurate according to the known rules governing that period of time, and you can count on it being correct, even if unintuitive.
- Generalizing units of time, such as years being 365 days always, or months being 30 days always, will cause problems, period. The bugs that result are far more likely to cause problems, than being surprised that two dates are not as far apart or close together as you assume at first glance.
- I will always favor correctness over consistency - time is dirty, messy, and all but impossible to reason about, assuming we can tame that is a fools errand. We can either pretend that an idealized view of time works for everyone (all too common), or accept the inherent mess, and focus on making sure the implementation is 100% accurate.

So with the above in mind, I'm going to revisit the current diffing/shifting and make sure it adheres to those goals, but I want to make sure that we're all on the same page as to what that means. For example, if you shift by 3 months, and then do a diff in terms of months, the results should be the same, however shifting by 3 months won't always mean the same thing (i.e. shifting across February is fewer days than shifting across March). The shifting/diffing should agree on what the period of time is, but it won't be the idealized view of that period.

Sound good?

gmile:
I just stumbled upon this issue when calculating a diff:

```
iex(1)> Timex.diff(~D[2018-10-15], ~D[2000-10-16], :years)
18
```
To me the result appears to be incorrect. I [checked](https://www.timeanddate.com/date/durationresult.html?d1=16&m1=10&y1=2000&d2=15&m2=10&y2=2018) against a random time comparison service on the internet. Given that service's result & Timex result I got this:

* number of days match:

    ```elixir
    > Timex.diff(~D[2018-10-15], ~D[2000-10-16], :days)
    6573
    ```

* number of years does not match:

    ```elixir
    Timex.diff(~D[2018-10-15], ~D[2000-10-16], :years)
    18
    ```

dovadi:
@gmile in order to avoid any confusion I like to say that with the fix in this pull request the result of the diff:

```Elixir
Timex.diff(~D[2018-10-15], ~D[2000-10-16], :years)
```

is 17 as you would expect.

bitwalker:
I've pushed the changes I described previously, which should address the issues in this PR. Can you all give it a look and let me know if you find any edge cases I haven't? I'm going to close this PR for now, so please open new issues if you find further problems.

gmile:
@bitwalker thank you! this I problem is fixed now:

```
Timex.diff(~D[2018-10-15], ~D[2000-10-16], :years)
```

PR #291, opened 2017-03-24
Take open and closed ends of Interval into consideration when checking membership

fsabe:
### Summary of changes

The Interval `member?` does not take into consideration the open and closed ends (`left_open` and `right_open` fields). On top of that, the current implementation excludes the left end of the interval despite including the right end (this is the opposite of the defaults).
 
```elixir
iex(1)> ~D[2017-03-24] in Timex.Interval.new(from: ~D[2017-03-24], until: ~D[2017-03-26])
false
iex(2)> ~D[2017-03-26] in Timex.Interval.new(from: ~D[2017-03-24], until: ~D[2017-03-26])
true
iex(5)> ~D[2017-03-26] in Timex.Interval.new(from: ~D[2017-03-24], until: ~D[2017-03-26], right_open: true)
true
```

This changes the implementation to take the right and left ends into consideration when checking membership.

### Checklist

### DIFF
--- a/lib/interval/interval.ex
+++ b/lib/interval/interval.ex
@@ -194,15 +194,20 @@ defmodule Timex.Interval do
       do_reduce({get_starting_date(interval), interval.until, interval.right_open, interval.step}, acc, fun)
     end
 
-    def member?(%Timex.Interval{from: from, until: until}, value) do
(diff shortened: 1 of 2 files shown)


bitwalker:
Looks good! Great PR :)

PR #293, opened 2017-03-29
Correct typing for add/2, subtract/2, shift/2

paulswartz:
### Summary of changes

The documentation (correctly) specifies that `shift/2` can return an AmbiguousDateTime struct for various reason.  However, that isn't a `Types.valid_datetime` so it doesn't correctly pass dialyzer checks.  Since `add/2` and `subtract/2` call `shift/2` as their implementation, they can also return those values.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -1375,7 +1375,7 @@ defmodule Timex do
   Same as `shift(date, Duration.from_minutes(5), :duration)`.
   """
   @spec add(Convertable.t, Duration.t) ::
-    Date.t | NaiveDateTime.t | DateTime.t | {:error, term}
+    Types.valid_datetime | AmbiguousDateTime.t | {:error, term}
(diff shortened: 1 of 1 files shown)


lib/timex.ex
```diff
@@ -1375,15 +1375,16 @@ defmodule Timex do
   Same as `shift(date, Duration.from_minutes(5), :duration)`.
   """
   @spec add(Convertable.t, Duration.t) ::
-    Date.t | NaiveDateTime.t | DateTime.t | {:error, term}
+  Types.valid_datetime | | AmbiguousDateTime.t | {:error, term}
```
bitwalker: You have a typo here (two `|`), which is breaking the build. Could you fix that up?

paulswartz: Sorry about that! should be better now.

bitwalker:
Thanks again for the PR!

PR #296, opened 2017-04-03
Cause Interval reduce to halt with invalid step

### LINKED ISSUE #292: Interval with incorrect step leads to infinite cycle on enumerating
Steps to reproduce:

```elixir
Interval.new(step: [month: 1]) |> Enum.count 
```

I guess this is due to lack of check in `Interval.has_recursion_ended?/3`:

```elixir
defp do_reduce({current_date, end_date, right_open, keywords}, {:cont, acc}, fun) do
  if has_recursion_ended?(current_date, end_date, right_open) do
    {:done, acc}
  else
    # here .shift returns {:error, {:unknown_shift_unit, :month}}
    # which is then passed to .has_recursion_ended? without a check
    next_date = Timex.shift(current_date, keywords)
    do_reduce({next_date, end_date, right_open, keywords}, fun.(current_date, acc), fun)
  end
end
```

Had hard time figuring this out; accidentally typing `month` instead of `months` drove me crazy.

I guess we can add the check to `Interval.new` and `Interval.with_step`, raising `FormatError` on incorrect steps. Can tackle this if suggested solution is acceptable.

mbramson:
Without halting when the step field was not a valid field for use in
Timex.shift, any function in Enum that relied on the Interval's
implementation of Enum.reduce would infinitely recurse.

Add validation to step unit for Interval.new/1

Add validation to step unit for Interval.with_step/2

Closes #292 

### Summary of changes

### Checklist

### DIFF
--- a/lib/interval/interval.ex
+++ b/lib/interval/interval.ex
@@ -15,6 +15,15 @@ defmodule Timex.Interval do
     end
   end
 
+  @type t :: %__MODULE__{}
+  @type valid_interval_step :: [microseconds: integer()] |
+                               [milliseconds: integer()] |
+                               [seconds:      integer()] |
+                               [minutes:      integer()] |
+                               [hours:        integer()] |
+                               [days:         integer()] |
+                               [weeks:        integer()]
+
   @enforce_keys [:from, :until]
   defstruct from:       nil,
             until:      nil,
@@ -69,6 +78,7 @@ defmodule Timex.Interval do
       "[15:30, 15:50]"
 
   """
+  @spec new(Keyword.t) :: Interval.t | {:error, :invalid_until} | {:error, :invalid_step}
   def new(options \\ []) do
     from = case Keyword.get(options, :from) do
       nil -> Timex.Protocol.NaiveDateTime.now()
@@ -86,14 +96,29 @@ defmodule Timex.Interval do
               %Date{} = d          -> Timex.to_naive_datetime(d)
               _ -> {:error, :invalid_until}
             end
-    case until do
-      {:error, _} = err -> err
-      _ ->
-        %__MODULE__{from: from, until: until,
-                    left_open: left_open, right_open: right_open,
-                    step: step}
-    end
+
+    attrs = %{from: from,
+              until: until,
+              left_open: left_open,
+              right_open: right_open,
+              step: valid_step_or_error(step)}
+    build_struct_or_error(attrs)
   end
+
+  @spec build_struct_or_error(map()) :: Interval.t | {:error, atom()}
+  defp build_struct_or_error(%{until: err = {:error, _}}), do: err
+  defp build_struct_or_error(%{step:  err = {:error, _}}), do: err
+  defp build_struct_or_error(valid_attrs),                 do: struct(__MODULE__, valid_attrs)
+
+  @spec valid_step_or_error(any()) :: valid_interval_step | {:error, :invalid_step}
+  defp valid_step_or_error(step = [milliseconds: _]), do: step
+  defp valid_step_or_error(step = [microseconds: _]), do: step
+  defp valid_step_or_error(step = [seconds: _]),      do: step
+  defp valid_step_or_error(step = [minutes: _]),      do: step
+  defp valid_step_or_error(step = [hours: _]),        do: step
+  defp valid_step_or_error(step = [days: _]),         do: step
+  defp valid_step_or_error(step = [weeks: _]),        do: step
+  defp valid_step_or_error(_),                        do: {:error, :invalid_step}
 
   @doc """
   Return the interval duration, given a unit.
@@ -145,8 +170,12 @@ defmodule Timex.Interval do
       ["2014-09-22", "2014-09-25"]
 
   """
+  @spec with_step(Interval.t, any()) :: Interval.t | {:error, :invalid_step}
   def with_step(%__MODULE__{} = interval, step) do
-    %__MODULE__{interval | :step => step}
+    case valid_step_or_error(step) do
+      error = {:error, _} -> error
+      step                -> %__MODULE__{interval | step: step}
+    end
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/interval/interval.ex
```diff
@@ -145,8 +162,12 @@ defmodule Timex.Interval do
       ["2014-09-22", "2014-09-25"]
 
   """
+  @spec with_step(Interval.t, any()) :: :ok | {:errpr, :invalid_step}
```
hedgesky: I guess there is a typo here: `:errpr`.

lib/interval/interval.ex
```diff
@@ -145,8 +162,12 @@ defmodule Timex.Interval do
...
   def with_step(%__MODULE__{} = interval, step) do
-    %__MODULE__{interval | :step => step}
+    case validate_step(step) do
+      :ok -> %__MODULE__{interval | :step => step}
```
hedgesky: Maybe it's worth to use `step: step` syntax here to match style of the whole file?

lib/interval/interval.ex
mbramson: Good catch!

lib/interval/interval.ex
mbramson: Ah, I just kept the same line as before (with the validation around it). Good call!

lib/interval/interval.ex
```diff
@@ -89,12 +92,26 @@ defmodule Timex.Interval do
...
-        %__MODULE__{from: from, until: until,
-                    left_open: left_open, right_open: right_open,
-                    step: step}
+        case validate_step(step) do
+          :ok ->
+            %__MODULE__{from: from, until: until,
+                        left_open: left_open, right_open: right_open,
+                        step: step}
+          err -> err
+        end
```
mbramson: So I'll be he first person to say that I'm not in love with this nested case solution to this problem. Was attempting to chain a new validate_until and validate_step together, but it wasn't any more elegant than this. Any thoughts?

lib/interval/interval.ex
```diff
@@ -220,8 +241,12 @@ defmodule Timex.Interval do
...
         {:done, acc}
       else
-        next_date = Timex.shift(current_date, keywords)
-        do_reduce({next_date, end_date, right_open, keywords}, fun.(current_date, acc), fun)
+        case Timex.shift(current_date, keywords) do
+          {:error, _} ->
+            {:halt, acc}
+          next_date ->
+            do_reduce({next_date, end_date, right_open, keywords}, fun.(current_date, acc), fun)
+        end
```
mbramson: Even though `Timex.new/1` and `Timex.with_step/2` now return errors when given an invalid step, we still need to prevent infinite recursion here since you can just create an Interval struct by hand if you wanted.

lib/interval/interval.ex
hedgesky: What do you think about this solution?

```elixir
def new(options \\ []) do
  ...
  step_or_error = valid_step_or_error(step)
  attrs = %{ from: from, until: until, left_open: left_open, 
             right_open: right_open, step: step_or_error}
  build_struct_or_error(attrs)
end

# helper functions
defp valid_step_or_error(step) do
  case validate_step(step) do
    :ok -> step
    error -> error
  end
end

defp build_struct_or_error(%{ until: err = {:error, _} }), do: err
defp build_struct_or_error(%{ step:  err = {:error, _} }), do: err
defp build_struct_or_error(valid_attrs), do: struct(__MODULE__, valid_attrs)
```

mbramson: I like it. Definitely flatter.

mbramson:
@hedgesky @bitwalker I've done a bit of refactoring to clean up `Interval.new/1` as well as added number of additional tests. Let me know what you think!

test/interval_test.exs
```diff
@@ -30,6 +88,11 @@ defmodule IntervalTests do
     assert ["15:00", "15:10", "15:20", "15:30", "15:40", "15:50"] == steps
   end
 
+  test "halts when enumerating with an invalid step unit" do
+    interval = %Interval{from: ~D[2017-04-02], until: ~D[2017-05-02], step: [invalid_step: 1]}
+    assert 0 == Enum.count(interval)
```
hedgesky: I'd rather get an error raised here. From my point of view, returning 0 in this situation may be misleading and possibly hard to debug. "Fail fast", that sort of things. What do you think?

test/interval_test.exs
```diff
@@ -5,6 +5,64 @@ defmodule IntervalTests do
...
+      assert %Interval{step: [microseconds: 5]} = Interval.new(step: [microseconds: 5])
+      assert %Interval{step: [milliseconds: 5]} = Interval.new(step: [milliseconds: 5])
+      assert %Interval{step: [seconds: 5]}      = Interval.new(step: [seconds: 5])
+      assert %Interval{step: [minutes: 5]}      = Interval.new(step: [minutes: 5])
+      assert %Interval{step: [hours: 5]}        = Interval.new(step: [hours: 5])
+      assert %Interval{step: [days: 5]}         = Interval.new(step: [days: 5])
+      assert %Interval{step: [weeks: 5]}        = Interval.new(step: [weeks: 5])
+    end
+
+    test "returns an error tuple when given an invalid step" do
```
hedgesky: 👍

test/interval_test.exs
```diff
@@ -5,6 +5,64 @@ defmodule IntervalTests do
...
+      assert {:error, :invalid_step} = Interval.new(step: [invalid_step: 5])
+    end
```
hedgesky: 👍

test/interval_test.exs
mbramson: I thought about this. You're definitely right that having it just not iterate at all could be confusing. I'll make this change. Any thoughts on what an appropriate error type would be here?

hedgesky: I guess it could be already presented `Timex.Interval.FormatError` with specified message. For example, `message: "Invalid step option"`.

mbramson: Made this change. Let me know what you think!

lib/interval/interval.ex
```diff
@@ -220,8 +249,14 @@ defmodule Timex.Interval do
       if has_recursion_ended?(current_date, end_date, right_open) do
         {:done, acc}
       else
-        next_date = Timex.shift(current_date, keywords)
-        do_reduce({next_date, end_date, right_open, keywords}, fun.(current_date, acc), fun)
+        case Timex.shift(current_date, keywords) do
+          {:error, {:unknown_shift_unit, :invalid_step}} ->
```
hedgesky: We shouldn't use specific step value here; we'd better use ```{:error, {:unknown_shift_unit, _}} ->``` instead.
You could check it by using another step value in specs:

```elixir
interval = %Interval{from: ~D[2017-04-02], until: ~D[2017-05-02], step: [another: 1]}
assert_raise Interval.FormatError, "Invalid step unit for %Timex.Interval{}", fn ->
  Enum.count(interval)
end
```
It will fail.

mbramson: Duh. I forgot that :invalid_step was what I was testing for, and assumed it was just part of the error that Timex.shift regularly returns. Great catch.

hedgesky:
Now it looks great! 👍

bitwalker:
This looks great! If you are ready for me to merge, I'll go ahead and do so :)

mbramson:
@bitwalker Yep, I think this is good to go!

bitwalker:
Thanks for the great work and your patience!

PR #303, opened 2017-05-07
Fix pattern matching in Timex.from_iso_triplet/3 error clause

hedgesky:
### Removed redundant `Timex.from_iso_triplet/3`

There isn't "useful" clause for `Timex.from_iso_triplet/3`, only an error one; so I guess we can safely remove it.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -633,7 +633,7 @@ defmodule Timex do
       {year, month, day} = Helpers.iso_day_to_date_tuple(year, iso_day)
       %Date{year: year, month: month, day: day}
   end
-  def from_iso_triplet(_, _, _), do: {:error, {:from_iso_triplet, :invalid_triplet}}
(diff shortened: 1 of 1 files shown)


lib/timex.ex
```diff
@@ -633,7 +633,6 @@ defmodule Timex do
       {year, month, day} = Helpers.iso_day_to_date_tuple(year, iso_day)
       %Date{year: year, month: month, day: day}
   end
-  def from_iso_triplet(_, _, _), do: {:error, {:from_iso_triplet, :invalid_triplet}}
```
bitwalker: It should be `def from_iso_triplet({_,_,_}), do: {:error, ...}`, that's why it appears that there is no useful clause. Could you update this PR with that change? I'll merge it once it is. Thanks!

hedgesky: Yeah, definitely. Missed guard clause in the upper clause, thought it would catch all triplets. Done.

bitwalker:
Thanks for the PR!

PR #304, opened 2017-05-08
The full name for May in Italian is Maggio

wstucco:
Just a minor typo in the translation files: the full name for May in Italian is Maggio


bitwalker:
Thanks!

Issue #302, opened 2017-04-26
RFC: Timex.equal? – Be explicit about comparison granularity


fpoehler:
Example

```
a = DateTime.from_naive! ~N[2017-04-26 08:08:08.123Z], "Etc/UTC"
b = DateTime.from_naive! ~N[2017-04-26 08:08:08.456Z], "Etc/UTC"
Elixir.Timex.equal?(a, b)
=> true    # ... shouldn't this be false?
           # a and b are obviously _not_ equal,
           # 123 milliseconds != 456 milliseconds!
```

According to the documentation, the arguments of `Timex.equal?` are equal (only) if they "represent the same moment in time", just as one would expect.

However, the implementation only does a comparison with a `:seconds` granularity, i.e. two `DateTime`s having the same calendar, timezone, date, hour, minute and second, but different milliseconds, are still equal as of `Timex.equal?`, even though in reality they are not *equal*, but at best *looking mostly equal*.

This was not the behaviour I did expect after I have read the docs. "The same moment in time" is not the same as "the same second". – It's "the same second, same millisecond, same nanosecond and so on, down to the smallest representable unit of time".

To remove that ambiguity, I would like to propose

* either to mention in the documentation that `equal?` does not consider the sub-second parts of a `DateTime`, and that users who need sub-second granularity can fallback on `Timex.compare(a, b) == 0` as an alternative
* or (even better) to make the granularity configurable in the code (and to reflect that in the documentation):
    ```
    def equal?(a, b, granularity \\ :seconds) do
      Comparable.compare(a, b, granularity) == 0
    end
    ```

Thanks for considering.

bitwalker:
The latter of these two options is what should be there, I'll make sure this and the docs are updated. Thanks for the report!

PR #310, opened 2017-05-25
Add granularity option to Timex.equal?

alvinlindstam:
### Summary of changes

Fixes https://github.com/bitwalker/timex/issues/302 as suggested in the issue.

I'n not sure if `:seconds` is the correct default option, since compare and diff uses `:microseconds` as the default granularity. But using seconds means we don't change the current behaviour.

Also fixes a documentation error for compare/3. It was a mismatch between Timex.compare and Timex.Comparable.compare.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -771,8 +771,11 @@ defmodule Timex do
 
   @doc """
   Returns a boolean indicating whether the two `Timex.Comparable` values are equivalent.
-  Equality here implies that the two Comparables represent the same moment in time,
-  not equality of the data structure.
+
+  Equality here implies that the two Comparables represent the same moment in time (with
+  the given granularity), not equality of the data structure.
+
+  The options for granularity is the same as for `compare/3`, defaults to `:seconds`.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


lib/timex.ex
```diff
@@ -786,9 +789,10 @@ defmodule Timex do
       ...> #{__MODULE__}.equal?(date1, date2)
       true
   """
-  @spec equal?(Comparable.comparable, Comparable.comparable) :: boolean | {:error, :badarg}
-  def equal?(a, a), do: true
-  def equal?(a, b), do: Comparable.compare(a, b, :seconds) == 0
+  @spec equal?(Comparable.comparable, Comparable.comparable, Comparable.granularity) :: boolean | {:error, :badarg}
+  def equal?(a, a, granularity\\:seconds)
```
bitwalker: A nitpick, but could you put spaces around the `\\` operator?

alvinlindstam:
@bitwalker: I have fixed the code style issue (spacing around `\\`) and rebased it into the first commit.

bitwalker:
Thanks!

PR #314, opened 2017-06-16
Fixed month_shortname/1 Examples

lpedromachado:
### Summary of changes

### Checklist


bitwalker:
Thanks! Good catch!

PR #315, opened 2017-06-16
Fix typespecs for beginning_of_week and friends

randycoulman:
### Summary of changes

When running dialyzer against [my project](https://github.com/randycoulman/invoice_tracker), I discovered that the type spec for `Timex.beginning_of_week/2` was not correct.  The docs state:

> The weekstart can between 1..7, an atom e.g. :mon, or a string e.g. “Monday”

But the typespec only allows for `Timex.Types.weekday`, which is defined to be `non_neg_integer`.

Further investigation indicates that the same issue affects `end_of_week`, `days_to_beginning_of_week` and `days_to_end_of_week` as well.

### Checklist

### Details

There's not really a way to write tests for this that I can think of, but before making my changes, running `mix dialyze` reports a number of warnings, including:

```
lib/timex.ex:1286: Function days_to_end_of_week/1 has no local return

lib/timex.ex:1286: The call 'Elixir.Timex':days_to_end_of_week(x0@1::any(),'mon')
will never return since the success typing is ({{integer(),integer(),integer()},{byte(),byte(),byte()}} |
{integer(),integer(),integer()} | #{'__struct__':='Elixir.Date' | 'Elixir.DateTime' | 'Elixir.NaiveDateTime',
'calendar':=atom(), 'day':=integer(), 'month':=integer(), 'year':=integer(), 'hour'=>byte(),
'microsecond'=>{char(),0 | 1 | 2 | 3 | 4 | 5 | 6}, 'minute'=>byte(), 'second'=>byte(),
'std_offset'=>integer(), 'time_zone'=>binary(), 'utc_offset'=>integer(),
'zone_abbr'=>binary()},non_neg_integer()) -> non_neg_integer() | {'error',_} and the contract is
('Elixir.Timex.Types':valid_datetime(),'Elixir.Timex.Types':weekday()) -> integer() | {'error',term()}
```
After making these changes, those two warnings are gone.

Similarly, in my project I was getting these warnings for this code:

```elixir
def for_current_status(today \\ local_today()) do
  Timex.beginning_of_week(today, :fri)
end
```

```
lib/invoice_tracker/default_date.ex:37: Function for_current_status/0 has no local return
lib/invoice_tracker/default_date.ex:37: Function for_current_status/1 has no local return
lib/invoice_tracker/default_date.ex:39: The call 'Elixir.Timex':beginning_of_week(today@1::any(),'fri')
breaks the contract ('Elixir.Timex.Types':valid_datetime(),'Elixir.Timex.Types':weekday()) ->
'Elixir.Timex.Types':valid_datetime() | {'error',term()}
```
Those warnings are now gone with these changes.

An interesting side note: `mix dialyze` is only complaining about `days_to_end_of_week` and not `days_to_beginning_of_week`.  That's because `days_to_end_of_week` uses `:mon` as its default weekstart value, whereas `days_to_beginning_of_week` uses `1` instead.  I didn't change this to be consistent, though I could if that's desirable.

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -51,7 +51,7 @@ defimpl Timex.Protocol, for: Date do
   @spec end_of_day(Date.t) :: Date.t
   def end_of_day(%Date{} = date), do: date
 
-  @spec beginning_of_week(Date.t, Types.weekday) :: Date.t
+  @spec beginning_of_week(Date.t, Types.weekstart) :: Date.t
   def beginning_of_week(%Date{} = date, weekstart) do
     case Timex.days_to_beginning_of_week(date, weekstart) do
       {:error, _} = err -> err
@@ -59,7 +59,7 @@ defimpl Timex.Protocol, for: Date do
     end
   end
 
-  @spec end_of_week(Date.t, Types.weekday) :: Date.t
+  @spec end_of_week(Date.t, Types.weekstart) :: Date.t
   def end_of_week(%Date{} = date, weekstart) do
     case Timex.days_to_end_of_week(date, weekstart) do
       {:error, _} = err -> err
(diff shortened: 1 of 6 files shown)


bitwalker:
Thanks!

Issue #319, opened 2017-06-23
strftime parser does not handle fractional part if it is equal to zero


mspanc:
### Steps to reproduce

```elixir
Timex.parse!("2017-06-15T12:42:20.000000", "%FT%T.%f", :strftime)
** (ArgumentError) argument error
              :erlang.binary_to_integer("")
      (timex) lib/parse/datetime/helpers.ex:70: Timex.Parse.DateTime.Helpers.parse_microseconds/1
    (combine) lib/combine/parsers/base.ex:65: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:578: Combine.Parsers.Base.label_impl/3
    (combine) lib/combine/parsers/base.ex:63: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:163: Combine.Parsers.Base.do_pipe/3
    (combine) lib/combine/parsers/base.ex:152: Combine.Parsers.Base.pipe_impl/3
    (combine) lib/combine.ex:52: Combine.parse/2
```

### Description of issue

I tried to implement workaround for #318 and it seems that strftime parser is also buggy. It does not handle fractional part properly if it's made out of zeros. If you change it at least to be `.000001` it works fine.

Using 3.1.17.

PR #320, opened 2017-06-23
Parse a string of zeroes as valid microseconds

alvinlindstam:
The precision should be the given number of zeroes.

Fixes https://github.com/bitwalker/timex/issues/319

### Additional note

This only fixes `parse_microseconds`. It seems that [`parse_milliseconds`](https://github.com/bitwalker/timex/blob/dfd230c2e1380f5e544404e75c7d71747204d56d/lib/parse/datetime/helpers.ex#L57) in the same module would also crash on a string of zeroes.

I didn't touch it since I didn't really understand how it is used.

### Checklist


bitwalker:
Thanks!

PR #321, opened 2017-06-25
Test, refactor and fix bugs in Timex.Convert.convert_map

alvinlindstam:
This bumps the test coverage of convert.ex to 100%.

In doing so, I also noticed and fixed a few bugs:

* If the input map missed hour date, the expected Date was wrapped in
  {:ok, date} tuple.
* If the input map missed both timezone and microseconds, the returned
  NaiveDateTime had a default microseconds of {0, 6}. If the timezone was
  declared, the default microsecond was {0, 0}.
* `Timex.Convert.convert_map` crashed when a microsecond was included,
  since it accepted or generated a `Calendar.microsecond` tuple while
  the `Timex.DateTime.Helpers.construct` function only accepted integer
  microseconds.

### Summary of changes

See commit message above.

Should there be type specs for internal, private, functions?

It looks like the typespecs for some functions using erlang datetime tuples is wrong. The `Types.time` only allows for `{ hour, minute, second }` while many functions such as `Timex.Helpers.construct` accepts `{ hour, minute, second, microsecond }`. Its only partially related to this PR (the use of four tuple times is distributed in the codebase), but it's worth mentioning. It also seems that the four tuple times are sometimes treated as milliseconds and sometimes as microseconds.

### Checklist

### DIFF
--- a/lib/datetime/helpers.ex
+++ b/lib/datetime/helpers.ex
@@ -57,6 +57,7 @@ defmodule Timex.DateTime.Helpers do
     end
   end
 
+  def construct_microseconds({us, p} = us_tuple) when is_integer(us) and is_integer(p), do: us_tuple
   def construct_microseconds(0), do: {0,0}
   def construct_microseconds(n), do: {n, precision(n)}
 
--- a/lib/convert/convert.ex
+++ b/lib/convert/convert.ex
@@ -18,11 +18,11 @@ defmodule Timex.Convert do
           not(is_nil(year)) and not(is_nil(month)) and not(is_nil(day)) ->
             case Map.get(datetime_map, :hour) do
               nil ->
-                Date.new(year, month, day)
+                with {:ok, date} <- Date.new(year, month, day), do: date
               hour ->
                 minute = Map.get(datetime_map, :minute, 0)
                 second = Map.get(datetime_map, :second, 0)
-                us     = Map.get(datetime_map, :microsecond, 0)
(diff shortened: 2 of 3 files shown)


lib/datetime/helpers.ex
```diff
@@ -57,6 +57,7 @@ defmodule Timex.DateTime.Helpers do
     end
   end
 
+  def construct_microseconds(us_tuple = {us, p}) when is_integer(us) and is_integer(p), do: us_tuple
```
bitwalker: I prefer to always have the pattern on the left, i.e. `{us, p} = us_tuple`

bitwalker:
This looks great! Thanks for looking deeper into this module! Regarding the use of 4 tuples - I know it's messy - unfortunately it's partially because the 4 tuple is not official in the standard library of Erlang (or Elixir), but rather something found in some popular community libs, namely Ecto. If you point out some of the areas you noticed problems in this respect, open an issue with some links and I'll dig in and try to sort out if it's a bug, inconsistency, or intentional which one we use. I think at this point we should probably assume that the 4-tuple contains microseconds, but I'll need to review the places which would need to be changed to ensure we're not breaking something unintentionally.

As for this PR, just the one tweak and I'll get it merged ASAP :)

alvinlindstam:
@bitwalker: Fixed the pattern order thingy.

I'm looking into the dialyzer warnings in general, will get back with more issues/PRs.

bitwalker:
Thanks!

PR #323, opened 2017-06-28
Always pad microseconds with zeroes to 6 digits

### LINKED ISSUE #322: strftime formatter does not pad microseconds when fractional part is equal to zero
### Steps to reproduce
```
iex(1)> Timex.to_datetime({2017, 1, 1}) |> Timex.format("%f", :strftime)
{:ok, "0"}
```
### Description of issue

Expected result: `{:ok, "000000"}`

For example, in Python:
```
>>> from datetime import datetime
>>> datetime.strftime(datetime(2017, 1, 1), '%f')
'000000'
```

narkq:
Fixes #322

### Summary of changes

Just as title says.
As I was upgrading from timex 2.2.1, strftime formatter stopped padding microseconds when fractional part was equal to zero.

### Checklist

### DIFF
--- a/lib/format/datetime/formatter.ex
+++ b/lib/format/datetime/formatter.ex
@@ -552,15 +552,11 @@ defmodule Timex.Format.DateTime.Formatter do
         pad_numeric(Timex.to_unix(date), flags, width)
     end
   end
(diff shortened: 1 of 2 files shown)


narkq:
It seems that I misinterpreted the documentation. See discussion in #322.

PR #325, opened 2017-07-01
Bugfix: Don't return bool on before? or after? on error

alvinlindstam:
### Summary of changes

Found this while fixing dialyzer issues. Comparable.compare might return an error tuple, which should not be used for sorting.

### Checklist


bitwalker:
Thanks for the PR!

PR #326, opened 2017-07-01
Bugfix: Fix match errors in Timex.Protocol implementation for tuples

alvinlindstam:
`d` was used both for the full date/datetime tuple and the day
item within it. It could never match.

### Checklist


bitwalker:
Nice catch!

PR #327, opened 2017-07-01
Bugfix: Should not crash when parsing ISO string with missing fractions

alvinlindstam:
### Summary of changes

Found this by fixing a warning from dialyzer

### Checklist


bitwalker:
Thanks!

PR #328, opened 2017-07-01
Add typespecs for Timex.Protocol and adjust irregularities

alvinlindstam:
Adding the typespecs to the protocol will enable dialyzer
to verify the typespecs of the implementations as well as
the calls from the main Timex module.

The typespecs in the Timex module was updated to use
Types.valid_datetime instead of the more verbose multi type
version (also Types.calendar_types does not exist).

Error tuple returns was added to those missing it in the
Timex module.

The protocol implementations type specifications is
narrowed down to match those of the protocol/caller module.

The typespecs for to_julian was changed to float, which is
what is returned.

Also fixes a few bugs:

* from_iso_day for Date now returns a Date struct instead of a map
* beginning_of_quarter for Date now works instead of crashes
* beginning_of_quarter now returns the correct time for erlang datetimes
* end_of_week now returns a erlang datetime instead of a date for erlang datetimes

### Summary of changes

I have som more typespec changes coming up, but think that this is a good start to build upon an merge separately.

This greatly improves dialyzers ability to find errors in the typespecs for all the implementations and calling code. A possible downside is if any client implements their own implementation of the protocol, in which case I think they will get dialyzer errors since their types are not the expected types. Is this a dealbreaker? 

### Checklist

### DIFF
--- a/lib/date/date.ex
+++ b/lib/date/date.ex
@@ -9,18 +9,18 @@ defimpl Timex.Protocol, for: Date do
 
   @epoch_seconds :calendar.datetime_to_gregorian_seconds({{1970,1,1},{0,0,0}})
 
-  @spec to_julian(Date.t) :: integer
+  @spec to_julian(Date.t) :: float
   def to_julian(%Date{:year => y, :month => m, :day => d}) do
     Timex.Calendar.Julian.julian_date(y, m, d)
   end
 
-  @spec to_gregorian_seconds(Date.t) :: integer
+  @spec to_gregorian_seconds(Date.t) :: non_neg_integer
   def to_gregorian_seconds(date), do: to_seconds(date, :zero)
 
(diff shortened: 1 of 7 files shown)


bitwalker:
The protocol was designed to make implementations for the various types cleaner, I would expect it to be very rare that someone would need to implement the protocol types other than those supported, especially since plain maps are supported.

This is great! Thanks for digging in and helping clean up the code! Timex has evolved a ton since the beginning, so there is unfortunately still cruft hanging around that was missed. One thing I want to introduce are property based tests to help fuzz test the API and surface any latent bugs, it's tricky with date/time because we'll need custom generators and I haven't figured out a good way to do shrinking yet.

alvinlindstam:
Yes, improved tests would be very nice.

Happy to help out.

PR #330, opened 2017-07-01
Clarify and make sure that four-tuple times are accepted

alvinlindstam:
The fourth element is treated as microseconds, either as a raw integer
or as an elixir `Calendar.microsecond` tuple.

Adds helper functions to make it easier to implement all the erlang
compliant functions.

As mentioned in https://github.com/bitwalker/timex/pull/321, we don't have typespecs for tuples containing microseconds, although most/some of the tuple implementation functions accept it.

This is fairly opinionated, and explicitly allows microseconds in all tuple functions for the Timex protocol.

I opted for allowing both integer microseconds and microsecond tuples, since it was easy and since that is what we accept to the construct-function (in the helpers) and in the map implementation.

### Checklist

### DIFF
--- a/lib/datetime/erlang.ex
+++ b/lib/datetime/erlang.ex
@@ -4,50 +4,37 @@ defimpl Timex.Protocol, for: Tuple do
 
   @epoch :calendar.datetime_to_gregorian_seconds({{1970,1,1},{0,0,0}})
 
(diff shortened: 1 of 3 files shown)


bitwalker:
Mind rebasing on master? Looks like there is a conflict with one of your other PRs :)

alvinlindstam:
@bitwalker: Done

bitwalker:
Thanks!

PR #332, opened 2017-07-03
Dialyze fixes

alvinlindstam:
### Summary of changes

This is the last in my series of PRs from an effort to fix all dialyzer warnings. After this, only four warnings remain (I did not find a suitable solution for them):

> lib/l10n/gettext.ex:1: The pattern #{} can never match since previous clauses completely covered the type #{'count':=_, _=>_}
>
> lib/parse/datetime/parser.ex:92: The call 'Elixir.Timex':to_naive_datetime(dt@1::#{'__struct__':='Elixir.DateTime', 'time_zone':='nil', _=>_}) breaks the contract ('Elixir.Timex.Types':valid_datetime()) -> 'Elixir.NaiveDateTime':t() | {'error',term()}
> 
> lib/timezone/timezone.ex:288: Overloaded contract for 'Elixir.Timex.Timezone':resolve/3 has overlapping domains; such contracts are currently unsupported and are simply ignored
> 
> lib/timezone/timezone.ex:398: Overloaded contract for 'Elixir.Timex.Timezone':convert/2 has overlapping domains; such contracts are currently unsupported and are simply ignored

Ideally, I'd like there to be no dialyzer warnings and for the CI checks to fail or warn if they appear in PRs.

### Checklist

I did not squash the commits, since I think they would help in explaining the changes. There are some motivations for the changes in the commit messages.


bitwalker:
I'll take a look at the remaining dialyzer warnings and see if I can find a solution, then we can update the Travis config like you mentioned. Nice work!

PR #337, opened 2017-07-10
Partially fix incorrect typespecs

take-five:
This is attempt to fix warnings that Dialyzer emits in this project. Initially I tried to upgrade one of our projects to OTP 20, but after upgrade Dialyzer started complaining about `unknown type 'Elixir.Map'.t` (we run Dialyzer with flag `-Wunknown`). I found few occurrences of `Map.t` in dependencies, one of which is Timex.

So I cloned the repo and ran `mix dialyze --unknown` and got a red screen of warnings :) I've fixed few of them, but more emerged. Unfortunately, I don't know the codebase of `timex` too well to fix all the warnings, but I hope I made a step in right direction.

In its current state, Dialyzer emits 6 warnings without `-Wunknown` flag and 26 warnings with `-Wunknown` flag.

### DIFF
--- a/lib/convert/convert.ex
+++ b/lib/convert/convert.ex
@@ -5,7 +5,7 @@ defmodule Timex.Convert do
   Converts a map to a Date, NaiveDateTime or DateTime, depending on the amount
   of date/time information in the map.
   """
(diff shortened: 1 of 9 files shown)


bitwalker:
Looks great! Thanks!

PR #338, opened 2017-07-11
adds support for month and year steps in Timex.Interval

Musashi178:
adds support for month and year as a valid step for the Timex.Interval,
both in the “new” and “with_step” function. Also adds the given tests
into interval_test to make sure that the step type is correctly created.

### Summary of changes

The documentation of _Timex.Interval.with_step_ states "The step should be a keyword list valid for use with Timex.Date.shift.", but does not not support having a step size by months/years. This PR just adds those two step sizes

### Checklist


bitwalker:
Thanks!

Issue #340, opened 2017-07-12
RFC: Consider inclusive bounds for Timex.between?/3


xavier:
Right now `Timex.between?/3` is implemented by passing the lower and upper bounds as-is to `Timex.before?/2` and `Timex.after?/2` which are both bound exclusive (which is good... most of the time) 😄
This exclusive bound behaviour doesn't fit well with "between" semantics. For instance, the `BETWEEN` clause in SQL is inclusive and so are ranges in Elixir. I think that making the bounds inclusive would make the use of this function more intuitive and avoid many potential off-by-1 errors.

I'll be happy to work on this if you think that this might be a change for the best.
I understand the breaking nature of this change and I'll welcome any guidance as how to deal with it.

bitwalker:
I agree 100%, if you'd like to take a crack at it, that'd be great! As far as how to handle the fact that it's a breaking change, I think the best way to handle it is to introduce a new options parameter. The only option would be `inclusive: true | false`, with the default being `inclusive: false`. This gives us a deprecation cycle to change the default to `inclusive: true`, and either leave it at that or eventually deprecate the extra parameter with the new default behaviour.

As far as implementation, for the inclusive case I think we probably want to check if the input equals one of the bounds if neither before or after, using `equals?/3`, where the granularity is `:microseconds` unless another option `granularity: <unit>` is provided to override the comparison granularity.

Let me know what you think!

xavier:
All of this sounds great, I'll get to work.

Btw, when you say `equals?/3`, are you sure you didn't mean `compare/3`?
That would have been my go-to function to implement inclusive bound checking.

bitwalker:
@xavier Yes, use `compare/3` :)

PR #341, opened 2017-07-13
Add comparison options to Timex.between?

xavier:
### Summary of changes

Adds the `:inclusive` and  `:granularity` options to `Timex.between?`, as discussed in #340.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -765,17 +765,29 @@ defmodule Timex do
   end
 
   @doc """
-  Returns a boolean indicating whether the first `Timex.Comparable` occurs between the second and third
+  Returns a boolean indicating whether the first `Timex.Comparable` occurs between the second
+  and third.
+
+  By default, the `start`and `ending` bounds are *exclusive*. You can opt for inclusive bounds
+  by setting the `:inclusive` option to `true`.
+
   """
-  @spec between?(Comparable.comparable, Comparable.comparable, Comparable.comparable) ::
+  @type between_options :: [
+    inclusive: boolean
+  ]
+  @spec between?(Comparable.comparable, Comparable.comparable, Comparable.comparable, between_options) ::
     boolean | {:error, term}
-  def between?(a, start, ending) do
-    is_after_start? = after?(a, start)
-    is_before_end?  = before?(a, ending)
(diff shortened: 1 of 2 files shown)


alvinlindstam:
I'm not totally sure about including the `granularity` option. It would give a few strange results such as (strange as that the first date is not between the two limits, it's within at most a year from the limits, but that's not between them)

```elixir
> Timex.between? ~D[2018-03-01], ~D[2017-01-01], ~D[2017-06-30], :years
true
```

In what situations would one want to use a different option than microseconds?

If we do include it, shouldn't it also be supported for `after?` and `before?`?

xavier:
I reckon your `:years` example is problematic edge case but it's bound to the semantics of `compare/3`.
It might seem wrong at first that `Timex.compare(~D[2018-03-01], ~D[2017-06-30], :years) == 0` because the documentation for the `Comparable` protocol does not make it clear that the granularity applies to a distance of time rather than an actual comparison scope.

We could perhaps improve the documentation first and provide examples of potential counter-intuitive results to improve the developer experience.

bitwalker:
@alvinlindstam You bring up a good point, and I'm inclined to agree with you. It was my suggestion that we allow one to specify the comparison granularity for equality, but given the ambiguity of how the granularity is used it probably is best to use the most precise unit for comparison here. Documentation could address this, but the larger issue in my mind is that it leaves room for these kind of paradoxical comparisons where `between?(today, yesterday, tomorrow, granularity: :months)` returns false, when you'd expect true.

@xavier I apologize for the change of heart, but could you modify your implementation to not take `:granularity`, and instead use `:microseconds` in the comparisons?

I'm certainly open to further discussion on this, but I'd like to preserve the logical purity of how `before?`, `after?`, and `between?` intuitively work; I don't know that there is a convincing argument for making it more flexible when you can use `shift`, `to_date` or something like `beginning_of_day` to modify the inputs to the desired granularity. It means a little more work, but the resulting code will be more explicit and easier to understand.

xavier:
@bitwalker I've rewritten the change to only allow the `:inclusive` option.

bitwalker:
🎉  thanks again!

PR #344, opened 2017-07-18
Add functionality to compare Times

### LINKED ISSUE #339: Implement Timex.Comparable for times
I think it would be very useful for Timex to be able to compare times and we can give Time structs as arguments to functions like `between?`, `after?` and `before?`.

TBK145:
I refactored `before?` and friends a little bit to always use `Timex.compare/{2,3}`, this way I only had to add a special clause for `Time` structs for those functions.

Closes #339


bitwalker:
Thanks!

PR #346, opened 2017-07-19
Use String.trim with binary instead of String.strip to fix warning

TBK145:
This fix remains compatible with Elixir 1.3, because it was introduced in that version.


bitwalker:
Thanks!

PR #347, opened 2017-07-26
Fix typo for DateTime inspect

jayjun:
### Summary of changes

### Checklist


lib/datetime/inspect.ex
```diff
@@ -14,7 +14,7 @@ end
 # Only provide Inspect for versions < 1.5.0-rc*
 if System.version |> Version.parse! |> Version.compare(Version.parse!("1.5.0-rc.1")) == :lt do
   defimpl Inspect, for: DateTime do
-    def instpect(datetime, %{:structs => false} = opts) do
+    def inspect(datetime, %{:structs => false} = opts) do
```
bitwalker: Actually you can remove this clause completely - it's never used - good catch!

jayjun: Sure, too easy!

bitwalker:
Thanks!

Issue #220, opened 2016-09-30
Timex.Timezone.convert gives wrong result date/tz sets resulting ambiguous timezones


zoldar:
### Steps to reproduce

```elixir
iex(14)> {:ok, datetime} =  NaiveDateTime.new(2016, 10, 30, 0, 0, 0)
{:ok, ~N[2016-10-30 00:00:00]}
iex(15)> Timex.Timezone.convert(datetime, "Europe/Amsterdam")       
#<Ambiguous(#<DateTime(2016-10-30T02:00:00+02:00 Europe/Amsterdam)> ~ #<DateTime(2016-10-27T14:00:00+01:00 Europe/Amsterdam)>)>
```

### Description of issue

Expected: `DateTime` in `after` being offset by 1 hour
Actual: `DateTime` in `after` being offest by 3600 minutes

The source of the problem seems to be here:

https://github.com/bitwalker/timex/blob/master/lib/timezone/timezone.ex#L316

```elixir
before_diff = diff(tz, before_tz)
{before_shifted, before_us} = do_shift(seconds_from_zeroyear, microsecs, :minutes, before_diff)
after_diff = diff(tz, after_tz)
{after_shifted, after_us} = do_shift(seconds_from_zeroyear, microsecs, :minutes, after_diff)
```

Times are offset by minutes using diffs expressed in seconds.

Issue #142, opened 2016-03-28
Problems with timezone conversions around the end of daylight saving time


mkummer:
### Steps to reproduce

I just stepped across several difficulties converting timestamps from my local time zone to UTC and back, when they are occurring around the end of daylight saving time (the change occurred Oct 25 2015, 01:00 UTC).
Trying to convert a datetime right after the 'extra' hour when daylight saving time was ending last fall from my local timezone (Zurich) into UTC fails:
```elixir
Interactive Elixir (1.2.0) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Timex.datetime({ {2015, 10, 25}, {03, 12, 34} }, "Europe/Zurich") |> Timex.Timezone.convert("UTC")
** (FunctionClauseError) no function clause matching in Timex.Timezone.resolve/2
    (timex) lib/timezone/timezone.ex:182: Timex.Timezone.resolve("UTC", {:error, :badarg})
    (timex) lib/timezone/timezone.ex:268: Timex.Timezone.convert/2
```
The hour from 02:00:00–02:59:59 occurs twice in succession, first in the daylight saving time zone, then in the normalized time zone. But any value after 03:00:00 is unique, so this conversion is supposed to work fine.
Even worse, picking a timestamp during this 'duplicated' hour results in an unresponsive iex:
```elixir
iex(1)> Timex.datetime({ {2015, 10, 25}, {02, 12, 34} }, "Europe/Zurich") |> Timex.Timezone.convert("UTC")
[iex freezes with high CPU load, have to break using Ctrl-C]
```
Going the opposite way reveals a different error:
```elixir
iex(1)> Timex.datetime({ {2015, 10, 25}, {00, 12, 34} }, "UTC") |> Timex.Timezone.convert("Europe/Zurich")
#<Ambiguous(#<DateTime(2015-10-25T02:12:34 Europe/Zurich (+02:00:00))> ~ #<DateTime(2015-10-25T01:12:34 Europe/Zurich (+01:00:00))>)>
```
This time in fact is NOT ambiguous; the only correct result is `#<DateTime(2015-10-25T02:12:34 Europe/Zurich (+02:00:00))>`.

`#<DateTime(2015-10-25T01:12:34 Europe/Zurich (+01:00:00))>` actually is not valid, as the +01:00 timezone does not kick in before 02:00 (+01:00)

Last but not least, let's try the conversion UTC-to-local one hour later:
```elixir
[...]

PR #353, opened 2017-08-04
Respect given TimezoneInfo in Timezone.convert/2

alvinlindstam:
Previously, the convert function checked the new wall time after the tz
offset was applied, resolved it by it's name and wall time and then
converted by that.

This caused a few bugs:

1. If the resulting wall time of the conversion based on the offset
   of the passed TimezoneInfo fell during an hour which is ambiguous or
   non existent time, the function would return an AmbiguousDateTime,
   even though the input data is not ambiguous.
2. If the function was called with an AmbiguousTimezoneInfo as the
   second argument, the same function was called with each potential
   TimezoneInfo. Because of issue 1 above, that could result in an
   AmbiguousDateTime which would then be nested in another
   AmbiguousDateTime.
3. A custom TimezoneInfo (which is supported, using
   TimezoneInfo.create), would not work with this function. If the
   name did not match any non custom timezone, this would return an
   error tuple. If the full name matched a built in (which would
   probably be an error in itself?) the result would be based on the
   built in TimezoneInfo.

### Discussion

I call 1 above a bug, even thought it's been somewhat discussed in https://github.com/bitwalker/timex/issues/142 and https://github.com/bitwalker/timex/issues/220 and is probably intentional.

In https://github.com/bitwalker/timex/issues/142#issuecomment-202480105, @bitwalker wrote:

> @mkummer I did find one assumption that is incorrect here:
> 
> > This time in fact is NOT ambiguous; the only correct result is `#<DateTime(2015-10-25T02:12:34 Europe/Zurich (+02:00:00))>`.
> 
> Actually that time is ambiguous according to the wall clock.

I disagree with @bitwalker's reasoning there. Yes, according to the wall clock, it is ambiguous wether `2015-10-25T02:12:34` in `Europe/Zurich` is with or without DST. However, the question that a timezone conversion function should answer is not whether or not that time is ambiguous, but what the the correct time in the target timezone is for the moment in time provided in some other timezone. The moment in time passed into the example (`Timex.datetime({ {2015, 10, 25}, {00, 12, 34} }, "UTC")`), it represents a single moment in time and there is a single DateTime in `Europe/Zurich` which accurately represents that same moment in time.

This is different from `Timex.to_datetime({ {2015, 10, 25}, {02, 12, 34} }, "Europe/Zurich")`, which should check the given wall time and assume that it is in the given timezone, but return an AmbiguousDateTime if there is more than one moments in time with that given wall time:

```
iex> Timex.to_datetime({ {2015, 10, 25}, {02, 12, 34} }, "Europe/Zurich")
#<Ambiguous(#<DateTime(2015-10-25T02:12:34+02:00 Europe/Zurich)> ~ #<DateTime(2015-10-25T02:12:34+01:00 Europe/Zurich)>)>
```

### Checklist

### DIFF
--- a/lib/timezone/timezone.ex
+++ b/lib/timezone/timezone.ex
@@ -295,8 +295,8 @@ defmodule Timex.Timezone do
 
   If an invalid zone name is provided, an error will be returned
   """
-  @spec resolve(String.t, non_neg_integer, :utc | :wall) :: TimezoneInfo.t |
-    AmbiguousTimezoneInfo.t | {:error, term}
+  @spec resolve(String.t, non_neg_integer, :utc) :: TimezoneInfo.t | {:error, term}
+  @spec resolve(String.t, non_neg_integer, :wall) :: TimezoneInfo.t | AmbiguousTimezoneInfo.t | {:error, term}
   def resolve(tzname, datetime, utc_or_wall \\ :wall)
 
   def resolve(name, seconds_from_zeroyear, utc_or_wall)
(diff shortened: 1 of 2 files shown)


alvinlindstam:
I would also note that I think that the main issue with the convert function is that it allows naive datetimes (and all other valid timex types), and treats is as a DateTime in utc. I would prefer if timezone conversions were only done on DateTimes that already have a timezone, and that conversions from naive to aware DateTimes was explicit.

This would, however, be an issue for the next major release (breaking change) so not really in scope to this PR.

lib/timezone/timezone.ex
```diff
@@ -406,63 +406,26 @@ defmodule Timex.Timezone do
...
-        after_dt = %DateTime{:year => y, :month => m, :day => d,
-                             :hour => h, :minute => mm, :second => s,
-                             :microsecond => after_us,
-                             :time_zone => after_tz.full_name,
-                             :zone_abbr => after_tz.abbreviation,
-                             :utc_offset => after_tz.offset_utc,
-                             :std_offset => after_tz.offset_std}
-        %AmbiguousDateTime{:before => before_dt, :after => after_dt}
-    end
+    {{y,m,d},{h,mm,s}} = :calendar.gregorian_seconds_to_datetime(seconds_from_zeroyear)
```
bitwalker: This doesn't account for cases where the shift is unambiguously in a different variant of the provided timezone, i.e. you want CST (America/Chicago) but when you shift the date/time by the offset, the time it actually represents is CDT (daylight savings time for the same zone).

Additionally, we should verify that when we get an AmbiguousTimezoneInfo, that the provided TimezoneInfo is present in one of the branches; if it is not, we should return the AmbiguousDateTime because we cannot make a choice. If it is present, then we can assume the intent of the developer is that they get a DateTime in the provided TimezoneInfo's zone.

alvinlindstam:
> It is ambiguous though - there is no dispute about that

There is no dispute about wether `2015-10-25T02:12:34` in Europe/Zurich is ambiguous when no dst info is given. But do you claim that it is ambiguous what time in Europe/Zurich that is the same time as `2015-10-25T00:12:34+00:00`? It is not about choosing a side of some ambiguity, it is about mapping one representation to a moment in time to another representation of that same moment. The two possible moment in time for `2015-10-25T02:12:34` in Europe/Zurich does not represent the same moment in time, and only one of them represents the same moment as `2015-10-25T00:12:34+00:00`

Is there any part of the following paragraph that you don't agree with?

1. At `2015-10-24 23:30:00Z`, the time in *Europe/Zurich* is `2015-10-25 01:30:00+02:00`.
2. An hour later, at `2015-10-25 00:30:09Z` the time in *Europe/Zurich* is `2015-10-25 02:30:00+02:00`. This is the first occurrence of the 02-hour that will repeat.
3. Another hour later, at `2015-10-25 01:30:09Z`, the time in *Europe/Zurich* is `2015-10-25 02:30:00+01:00`. This is the second occurrence of the 02-hour that has repeated.
4. Yet another hour later, at `2015-10-25 02:30:09Z`, the time in *Europe/Zurich* is `2015-10-25 03:30:00+01:00`.
5. `2015-10-25 02:30:00+01:00` is in no way a representation of the same moment in time as `2015-10-25 00:30:09Z`.
6. `Timex.Timezone.convert` is about transforming a representation of a moment in time from one timezone to another.
7. `Timex.Timezone.convert` should be commutative. That is, if I have a DateTime in utc, convert it to another timezone, then I should be able to convert it back to utc and get the same datetime as I started with.

Most of the other comments are all connected to position that we should return AmbiguousDateTimes when converting from one timezone to another, which I disagree with. So I'll mainly adress those when the main issue is resolved.

lib/timezone/timezone.ex
alvinlindstam: I'm not sure I'm following. But if you call `Timex.Timezone.convert(date, "America/Chicago")`, then the clause below will call `do_get("America/Chicago", date, :utc)` which will return the correct TimezoneInfo for that moment in time in America/Chicago.

Do you have an example of an input that represents your issue?

alvinlindstam:
For reference, the [`Calendar`](https://github.com/lau/calendar) library converts the time zone as I expect it to:

```elixir
iex(4)> Calendar.DateTime.from_erl!({{2015, 10, 25}, {00, 12, 34}}, "UTC") |> Calendar.DateTime.shift_zone!("Europe/Zurich")  
#<DateTime(2015-10-25T02:12:34+02:00 Europe/Zurich)>
iex(5)> Calendar.DateTime.from_erl!({{2015, 10, 25}, {01, 12, 34}}, "UTC") |> Calendar.DateTime.shift_zone!("Europe/Zurich")
#<DateTime(2015-10-25T02:12:34+01:00 Europe/Zurich)>
```

lib/timezone/timezone.ex
bitwalker: Right, calling with the timezone name works just fine (just recall though that you can get AmbiguousTimezoneInfo if the time falls during a DST transition), because the lookup always gets the valid TimezoneInfo for that date/time. But since `convert` can be passed a `TimezoneInfo` directly, we have to ensure that it is still a valid timezone for the provided DateTime. As an example, if the `TimezoneInfo` represents `America/Chicago` during CST, but the provided date/time in `America/Chicago` is actually in CDT after being shifted, the `TimezoneInfo` will be incorrect.

My second point is that if the `TimezoneInfo` passed to `convert` is present in one of the two possibilities of a resulting `AmbiguousTimezoneInfo` we get when doing resolution, we can automatically resolve the ambiguity because the desired `TimezoneInfo` is technically valid.

Does that help?

bitwalker:
The problem with the current code is that we are doing a shift to UTC, then using `:wall` clock for the timezone lookup. We should do the shift to UTC and use `:utc` clock, this will fix the problem that you see between Calendar and Timex in your given example.

bitwalker:
If you can review master, I pushed a fix for the bug I think is causing the discrepancy you are having trouble with. Both Timex and Calendar now produce the same results for your example.

bitwalker:
Ping :)

Issue #349, opened 2017-07-30
Timex.AmbiguousDateTime error


nicocaille:
Hello,
i have a date string that i am first converting to a NaiveDatetime

i tried with two date_string different, one is working, the other one is bringing an error i can't figure out

date_string |> NaiveDateTime.from_iso8601! |> Timex.local

```
iex(132)> date_string = "2017-02-18 22:27:34"
"2017-02-18 22:27:34"
iex(133)> date_string |> NaiveDateTime.from_iso8601! |> Timex.local
#<DateTime(2017-02-18T20:27:34-02:00 America/Sao_Paulo)>
```

If i get a date_string one hour later, i am getting the following error:

```
iex(134)> date_string = "2017-02-18 23:27:34"                      
"2017-02-18 23:27:34"
iex(135)> date_string |> NaiveDateTime.from_iso8601! |> Timex.local
** (KeyError) key :full_name not found in: #<Ambiguous(#<TimezoneInfo(America/Sao_Paulo - -02 (-02:00:00))> ~ #<TimezoneInfo(America/Sao_Paulo - -03 (-03:00:00))>)>
    (timex) lib/timex.ex:82: Timex.local/1
```

after running a few more tests, it seems that the time slot 23:00:00 -> 23:59:59 is raising a Timex error:

```
iex(141)> date_string = "2017-02-18 22:59:59"                      
"2017-02-18 22:59:59"
iex(142)> date_string |> NaiveDateTime.from_iso8601! |> Timex.local
#<DateTime(2017-02-18T20:59:59-02:00 America/Sao_Paulo)>
iex(143)> date_string = "2017-02-18 23:00:00"                      
"2017-02-18 23:00:00"
iex(144)> date_string |> NaiveDateTime.from_iso8601! |> Timex.local
** (KeyError) key :full_name not found in: #<Ambiguous(#<TimezoneInfo(America/Sao_Paulo - -02 (-02:00:00))> ~ #<TimezoneInfo(America/Sao_Paulo - -03 (-03:00:00))>)>
    (timex) lib/timex.ex:82: Timex.local/1
iex(144)> date_string = "2017-02-18 23:59:59"                      
"2017-02-18 23:59:59"
iex(145)> date_string |> NaiveDateTime.from_iso8601! |> Timex.local
[...]

PR #354, opened 2017-08-04
Fix issue #349: crash on `Timex.local` on input with ambiguous wall time

alvinlindstam:
Previously, the timezone checked for ambiguity on the input date before
converting it to the actual timezone, so the local function could get
an AmbiguousTimezoneInfo for dates that aren't actually ambiguous in
the target timezone.

As far as I can see, `Timex.local/1` is just syntactic sugar for
`Timex.Timezone.convert(date, :local)`, which is the solution I've used.

The tests on this is intended to check that a DateTime can be safely converted to the local timezone even if the input or output time matches an ambiguous wall time period. These tests will fail right now, since they depend on the changes I suggested in https://github.com/bitwalker/timex/pull/353. If that is rejected, I'll rewrite the tests.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -58,29 +58,14 @@ defmodule Timex do
   Returns a DateTime representing the current moment in time in the local timezone.
   """
   @spec local() :: DateTime.t | AmbiguousDateTime.t | {:error, term}
-  def local() do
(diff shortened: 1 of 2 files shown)


bitwalker:
I think this should be fine, but let's finish discussing/merging #353 first, then come back to this one.

PR #363, opened 2017-09-08
Fix #356 - parsing of milliseconds

### LINKED ISSUE #356: Parsing milliseconds (%L) results in ArgumentError for input "000"
Hi.

I think I may have found a bug: when parsing a millisecond field that is all-zeros using the :strftime formatter, Timex crashes with an ArgumentError.

The documentation states: "`%L` - milliseconds (000..999)", so I think 000 should be a valid value.

### Steps to reproduce

- `mix new <project>`
- Add Timex master branch as a dependency to mix.exs
- `iex -S mix`
```
Erlang/OTP 20 [erts-9.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

[...compiles app...]

iex(1)> Timex.parse!("2017-08-16T11:11:11.123", "%Y-%m-%dT%H:%M:%S.%L", :strftime)
~N[2017-08-16 11:11:11.123]
iex(2)> Timex.parse!("2017-08-16T11:11:11.000", "%Y-%m-%dT%H:%M:%S.%L", :strftime)
** (ArgumentError) argument error
    :erlang.binary_to_integer("")
    (timex) lib/parse/datetime/helpers.ex:58: Timex.Parse.DateTime.Helpers.parse_milliseconds/1
    (combine) lib/combine/parsers/base.ex:65: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:606: Combine.Parsers.Base.label_impl/3
    (combine) lib/combine/parsers/base.ex:63: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:163: Combine.Parsers.Base.do_pipe/3
    (combine) lib/combine/parsers/base.ex:152: Combine.Parsers.Base.pipe_impl/3
    (combine) lib/combine.ex:52: Combine.parse/3
iex(2)> Timex.parse!("001", "%L", :strftime)
~N[0000-01-01 00:00:00.001]
iex(3)> Timex.parse!("000", "%L", :strftime)
** (ArgumentError) argument error
    :erlang.binary_to_integer("")
    (timex) lib/parse/datetime/helpers.ex:58: Timex.Parse.DateTime.Helpers.parse_milliseconds/1
    (combine) lib/combine/parsers/base.ex:65: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:606: Combine.Parsers.Base.label_impl/3
    (combine) lib/combine/parsers/base.ex:63: Combine.Parsers.Base.map_impl/3
    (combine) lib/combine/parsers/base.ex:163: Combine.Parsers.Base.do_pipe/3
    (combine) lib/combine/parsers/base.ex:152: Combine.Parsers.Base.pipe_impl/3
    (combine) lib/combine.ex:52: Combine.parse/3

```

The code in question is this:
```elixir
def parse_milliseconds(ms) do
    n = ms |> String.trim("0") |> String.to_integer
    n = n * 1_000
    [sec_fractional: Timex.DateTime.Helpers.construct_microseconds(n)]
  end
```

`String.trim` empties the entire string, leaving the empty string for `String.to_integer`, which it doesn't like. (Edit: The code in question is also wrong for "010" because that is reduced to "1" and then parsed as 001 milliseconds.)

I could start to work on a pull request, but it's probably simpler to fix for a committer familiar with the project. Let me know which option you prefer.

DataWraith:
### Summary of changes

This fixes issue #356.

The use of `ms |> String.trim("0")` could result in an empty string when
the input was "000", which in turn made `String.to_integer` complain.

Since `trim` removes zeros from both ends of the string, milliseconds
ending in `0` or `00` were incorrectly parsed as well.

This commit changes `String.trim` to `String.trim_leading` and adds
a safety-clause to protect against the "000" input.

I tried to keep the changes minimal: I added a failing test to demonstrate the problem and then made it pass. I'm still very new to Elixir so the test is probably much uglier than it could be. I think `elem(x, 0)` is a smell here, but the second element of the tuple (the required precision) changed in ways I did not quite understand. I'd be happy to update the pull request with any needed changes.

I hope the pull request is acceptable without added typespecs -- still trying to figure those out.

### Checklist

### DIFF
--- a/lib/parse/datetime/helpers.ex
+++ b/lib/parse/datetime/helpers.ex
@@ -55,7 +55,8 @@ defmodule Timex.Parse.DateTime.Helpers do
   end
 
   def parse_milliseconds(ms) do
-    n = ms |> String.trim("0") |> String.to_integer
+    n = ms |> String.trim_leading("0")
+    n = if n == "", do: 0, else: String.to_integer(n)
     n = n * 1_000
     [sec_fractional: Timex.DateTime.Helpers.construct_microseconds(n)]
   end
--- a/test/parse_strftime_test.exs
+++ b/test/parse_strftime_test.exs
@@ -37,6 +37,16 @@ defmodule DateFormatTest.ParseStrftime do
     assert {:ok, ^date} = parse("20150713 14:01:21.038", "%Y%m%d %H:%M:%S.%L")
   end
 
+  test "issue #356 - millisecond parsing" do
+    {:ok, ms1} = parse("000", "%L")
+    {:ok, ms2} = parse("010", "%L")
+    {:ok, ms3} = parse("100", "%L")
+
+    assert elem(ms1.microsecond, 0) == 0
+    assert elem(ms2.microsecond, 0) == 10 * 1_000
+    assert elem(ms3.microsecond, 0) == 100 * 1_000
+  end
+
   defp parse(date, fmt) do
     Timex.parse(date, fmt, :strftime)
   end


bitwalker:
Thanks!

Issue #350, opened 2017-07-31
Week durations are not recognized


adamnbowen:
Timex.Duration doesn't seem to parse weeks in given durations.  "M" and other identifiers work as expected, but "W" does not:

```
iex(1)> Duration.parse("P1M")
{:ok, #<Duration(P1M)>}
iex(2)> Duration.parse("P1W")
{:error, "unexpected token W"}
```
As an example of another implementation of week parsing support, you can view my snippet running in with the node library iso8601-duration here: https://runkit.com/adamnbowen/597fa5d955331200125dbe02

I'm not quite sure where it's listed in the actual specification, but you can see the week format for durations mentioned in the Wikipedia entry on iso8601: https://en.wikipedia.org/wiki/ISO_8601#Durations

bitwalker:
I'm about to merge the PR which fixes this, so I'm going to close this issue, we're just wrapping up review.

PR #365, opened 2017-09-10
Add support for weeks in iso8601 duration parser (#350)

mlovic:
According to the ISO 8601, there are two formats for representing a duration:

> In both basic and extended format the complete representation of the expression for duration shall be [PnnW] or [PnnYnnMnnDTnnHnnMnnS].

This PR implements the first of those, as requested in #350.

    iex(1)> Timex.Duration.parse("P2W")
    {:ok, #<Duration(P14D)>}

However, I'm unsure about how to handle mixing the formats. While the ISO standard does not explicitly prohibit combining the two formats, it does clearly distinguish them, suggesting that they should be used separately. Because of this, I am inclined to treat any combination of the two formats as invalid input. However, both the Nodejs library linked to in the issue and Python's `isodate` (to give another example), _do_ support this, in different ways:

The node library will ignore the `W` designator, or everything except `W`, depending on which comes first.

The Python library will take into account both. From the library's test suite:

> ```
> (parse_duration("P1Y1M5W"),
>    [...should equal the difference between...]
>    date(2000, 1, 30),
>    date(2001, 4, 4)),
> ```

I also added some typespecs for the internal parsing functions in a different commit. It took me a while to understand those, so I thought the specs could be helpful. I'd be happy to remove them though.

### Checklist

### DIFF
--- a/lib/parse/duration/parsers/iso8601.ex
+++ b/lib/parse/duration/parsers/iso8601.ex
@@ -24,6 +24,10 @@ defmodule Timex.Parse.Duration.Parsers.ISO8601Parser do
       ...> Timex.Format.Duration.Formatter.format(d)
       "PT3H12M25.001S"
 
+      iex> {:ok, d} = #{__MODULE__}.parse("P2W")
+      ...> Timex.Format.Duration.Formatter.format(d)
+      "P14D"
+
       iex> #{__MODULE__}.parse("P15YT3D")
       {:error, "invalid use of date component after time separator"}
 
@@ -34,6 +38,8 @@ defmodule Timex.Parse.Duration.Parsers.ISO8601Parser do
     case parse_components(rest, []) do
       {:error, _} = err ->
         err
+      [{?W, w}] ->
+        {:ok, Duration.from_days(7*w)}
       components when is_list(components) ->
         result = Enum.reduce(components, {false, Duration.zero}, fn
           _, {:error, _} = err ->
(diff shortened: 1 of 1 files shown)


bitwalker:
This is great! Thanks for the PR!

The key is in the language used to describe the "basic" vs "extended" formats:

> In both basic and extended format the complete representation of the expression for duration shall be [PnnW] or [PnnYnnMnnDTnnHnnMnnS].

As somebody who is deep into parser theory and implementation, to me this reads that the "complete representation" of a duration in basic format is `PnnW`, and the "complete representation" of a duration in extended format is `PnnYnnMnnDTnnHnnMnnS` and that mixing the two should be considered invalid according to the spec (as implied by the "or"). I suspect Node's parser simply parses a "valid" format, either basic or extended, until it has fully parsed one of the two formats, ignoring what's left. Python's behaviour is interesting (even convenient!), but I believe it to be effectively an extension of the spec, and I would prefer to avoid going down that road in Timex.

lib/parse/duration/parsers/iso8601.ex
```diff
@@ -54,6 +60,8 @@ defmodule Timex.Parse.Duration.Parsers.ISO8601Parser do
             {true, Duration.add(d, Duration.from_minutes(m))}
           {?S, s}, {true, d} ->
             {true, Duration.add(d, Duration.from_seconds(s))}
+          {?W, _w}, {_, _} ->
+            {:error, "W cannot be used with other time period designators"}
```
bitwalker: Perhaps the error will be more helpful if we phrase it like:

```elixir
{?W, _w}, {_, _} ->
  {:error, "Found 'W', a basic format designator, but the parse indicates " <>
               "this is an extended format, mixing the two formats is disallowed"}
```

mlovic: :+1: I've amended the commit

bitwalker:
Thanks!

PR #372, opened 2017-09-26
Correct spurious Tuple.to_date/2

barbeque:
### Summary of changes

I noticed that the [Inch CI docs test page](https://inch-ci.org/github/bitwalker/timex) was reporting that `Timex.Protocol.Tuple.to_date/2` was missing documentation.

After hunting it down in `erlang.ex` it looks like it was a copy-paste mistake and was intended to be a fail-through guard case of `to_datetime/2`.

I added a test which I believe covers this guard case.

### Checklist


bitwalker:
Good catch!

PR #375, opened 2017-09-28
Fix #360 Update shift and diff by month

### LINKED ISSUE #360: Ensure shifting is consistent for both DateTime and Date/NaiveDateTime
### Steps to reproduce
```
iex(1)> date = ~D[2017-01-31]
~D[2017-01-31]
iex(2)> date |> Timex.shift(months: 1)
~D[2017-02-28]
iex(3)> date |> Timex.to_datetime |> Timex.shift(months: 1)
#DateTime<2017-03-03 00:00:00Z>
```
The last result should be: `#DateTime<2017-02-28 00:00:00Z>`

### Description of issue

- `shift()` should return the same result regardless of date format
- I prefer the way Timex implements the NavieDate shift.  When shifting to another month that is shorter than the original month, the shifted date should be adjusted to the end of the month
- The difference between the two shift behaviors is properly documented in:
https://github.com/bitwalker/timex/blob/master/lib/datetime/datetime.ex#L326-L327
https://github.com/bitwalker/timex/blob/master/lib/datetime/naivedatetime.ex#L257-L258
- It looks like the change to `shift` for DateTime was introduced while fixing https://github.com/bitwalker/timex/issues/313

Would you be willing to make these consistent?  
I can submit a pull request for the change.

tmaszk:
### Summary of changes

This PR addresses the issue raised in https://github.com/bitwalker/timex/issues/360 and the ensuing discussion.

It accomplishes the goal of making the month `shift` consistent between `datetime.ex` and `nativetime.ex` and also makes `shift` and `diff` by month mirrors of each other.

It also is consistent with the behavior of other date libraries I have used, like ruby's Date class, where shifting January 31 ahead three months is April 30.

`shift` examples:
```
iex(1)> Timex.shift ~D[2017-01-31], months: 1
~D[2017-02-28]
iex(2)> Timex.shift ~D[2017-01-31], months: 2
~D[2017-03-31]
iex(3)> Timex.shift ~D[2017-01-31], months: 3
~D[2017-04-30]
```
same thing in Ruby
```
irb(main):001:0> Date.new(2017, 1, 31) >> 1
=> #<Date: 2017-02-28 ((2457813j,0s,0n),+0s,2299161j)>
irb(main):002:0> Date.new(2017, 1, 31) >> 2
=> #<Date: 2017-03-31 ((2457844j,0s,0n),+0s,2299161j)>
irb(main):003:0> Date.new(2017, 1, 31) >> 3
=> #<Date: 2017-04-30 ((2457874j,0s,0n),+0s,2299161j)>
```

It does cause `diff` to be sometimes different depending on the order of the arguments, but that is consistent with the `shift` behavior, and consistent with what I would expect.
```
iex(1)> Timex.diff ~D[2017-01-31], ~D[2017-02-28], :months
-1
iex(2)> Timex.diff ~D[2017-02-28], ~D[2017-01-31], :months
0
iex(3)> Timex.diff ~D[2017-02-28], ~D[2017-01-28], :months
1
```

I realize that is a change is in contrast to the breaking change made in version 2.0.0 as noted by:  https://github.com/bitwalker/timex/blob/master/CHANGELOG.md#L263
```
BREAKING diff/3 now returns the same value no matter which order the arguments are given. Use compare/3 to get the ordering
```
I wouldn't suggest a breaking change unless I felt strongly that the new behavior is better than the prior behavior.

I didn't see the specific behavior of monthly `shift` or `diff` in the documentation, so I didn't submit any changes to it.  I can write it up if you think this is a good change and that it is of interest.

### Checklist

- [x] New functions have typespecs, changed functions were updated - no new functions
- [x] Same for documentation, including moduledocs - none for now, let me know if you want some
- [ ] Commits were squashed into a single coherent commit - I left it as two commits, one for `diff` and one for `shift`

### DIFF
--- a/lib/comparable/comparable.ex
+++ b/lib/comparable/comparable.ex
@@ -38,6 +38,17 @@ defprotocol Timex.Comparable do
     - {:error, reason}: when there was a problem comparing,
       perhaps due to a value being passed which is not a valid date/datetime
 
+  ## Examples
+
+      iex> use Timex
+      iex> date1 = ~D[2014-03-04]
+      iex> date2 = ~D[2015-03-04]
+      iex> Timex.compare(date1, date2, :years)
+      -1
+      iex> Timex.compare(date2, date1, :years)
+      1
+      iex> Timex.compare(date1, date1)
+      0
   """
   @spec compare(comparable, comparable, granularity) :: compare_result
   def compare(a, b, granularity \\ :microseconds)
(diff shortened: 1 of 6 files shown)


bitwalker:
I'm cool with this, I think we should probably document the rather unique behavior of diffing/shifting by months, since I think it is the only shift/diff unit which is ambiguous. Once you've got a note about that in the function docs for those functions (feel free to write it up in `diff` and link to it from `shift`), then I'll get this merged :)

tmaszk:
@bitwalker I added some docs and doctests.  Let me know what you think.

bitwalker:
Thanks!

Issue #368, opened 2017-09-16
Feature Request: New DateTime from a Date and a Time


ryharris:
It would be nice to have some function (possibly as arguments for `Timex.to_datetime`) where a date struct and a time struct can be given to create a new DateTime.

For example:

```elixir
iex(1)> Timex.to_datetime(~D[2017-09-16], ~T[08:00:00])
iex(2)> #DateTime<2017-09-16 08:00:00Z>
```

Would you consider this? Or is it out of the question for Timex?

barbeque:
It would even be nice if DateTime.set would accept a time struct for the `time` option instead of the current `{h, m, s}` tuple:

```ex
Timex.local |> Timex.set([time: ~T[09:25:00.00]])
```

Would have the same effect as:

```ex
Timex.local |> Timex.set([time: {9, 25, 0}])     
#<DateTime(2017-09-25T09:25:00.718045-06:00 America/Edmonton)>
```

Would you accept a pull request for a change along these lines?

bitwalker:
I would accept PRs for both of these requests :)

bitwalker:
@ryharris However in your request, I would prefer that it be a tuple of `{Date.t, Time.t}` rather than two arguments, as the existing protocol expects the first argument to be a date/time, and the second to be a timezone.

michalmuskala:
FWIW, this is acheivable with just standard library:

```elixir
DateTime.from_naive!(NaiveDateTime.new(date, time), "Etc/UTC")
```

PR #376, opened 2017-09-29
WIP: Add an overload for the 'time' option of Timex.set/2 that accepts a Time struct

barbeque:
### Summary of changes

I wanted to be able to use the `Time` struct instead of the h/m/s tuple for `Timex.set` when modifying a `DateTime`:
```ex
iex> to_change = Timex.to_datetime({{2017, 7, 21}, {1, 2, 3}})
iex> changed = Timex.set(to_change, [time: ~T[09:52:33.000]])
#<DateTime(2017-07-21T09:52:33Z Etc/UTC)>
```

As discussed in #368, this is nice to have for consistency as it's something I had been missing from Timex when I first used it.

There should probably be an equivalent option for `date` but I wanted to get early feedback on this change first to make sure that I am doing things "the right way" before getting too deep into it.

Open questions:
 * Should the tests really have gone in the `parse_default_test` file? I didn't see a set of tests dedicated solely to `set/2` so was unsure where to put them.
 * How should I handle possibly obliterating the `validate` option when I call `Timex.set` after converting the struct into a tuple? (datetime.ex:198)

### Checklist

### DIFF
--- a/lib/datetime/datetime.ex
+++ b/lib/datetime/datetime.ex
@@ -194,6 +194,8 @@ defimpl Timex.Protocol, for: DateTime do
             else
               %{result | :hour => h, :minute => m, :second => s}
             end
+          {:time, t} ->
+            Timex.set(result, [time: {t.hour, t.minute, t.second}])
           {:day, d} ->
             if validate? do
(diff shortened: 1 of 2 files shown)


bitwalker:
Looks good! Thanks!

PR #379, opened 2017-10-12
Fix Timex.Duration.diff/3 documentation of available units

aspett:
### Summary of changes

While I was reading the docs for Timex.Duration.diff/3, I noticed that `:days` as an available unit was missing, and hoped it was a mistake. Upon further checking, it seems it was; so I have added it to the documentation.

### Checklist


bitwalker:
Thanks!

PR #380, opened 2017-10-14
Fix typo and formats in README.md

jung-hunsoo:
Two things are changed.

- Fix typo
- Re-format for better readability


bitwalker:
Thanks!

PR #381, opened 2017-10-16
Remove unused diff_years method

treble37:
### Summary of changes

I think this method is never hit `defp diff_years(a, a), do: 0` in the current implementation, so let's remove it.

### Checklist

### DIFF
--- a/lib/comparable/diff.ex
+++ b/lib/comparable/diff.ex
@@ -67,7 +67,6 @@ defmodule Timex.Comparable.Diff do
   defp zero(:duration), do: Duration.zero
   defp zero(_type), do: 0
 
-  defp diff_years(a, a), do: 0
   defp diff_years(a, b) do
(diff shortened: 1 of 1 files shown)


bitwalker:
Why would it never be hit? If the two years are the same value, then the diff should be 0.

treble37:
Maybe I am misunderstanding something - I run in an IEx shell using the following:

```
t = ~D[2017-01-01]
t2 = ~D[2017-02-01]
Timex.Comparable.Diff.diff(t, t2, :years)
#=> 0
```

Because of line 26, `defp do_diff(a, a, type),      do: zero(type)`, wouldn't Date(times) with the same years pattern match on line 26 first and the removed line would never be hit.  Unless there's another path to diff_years that I am not seeing?

bitwalker:
I've merged this manually, thanks!

PR #383, opened 2017-10-19
Fixing months diffing when year is different

### LINKED ISSUE #382: Timex.diff/3 with :months granularity is wrong when the year is not the same
### Steps to reproduce

Is pretty simple to repeat:

```elixir
iex(1)> Timex.diff(~D[2017-12-20], ~D[2017-10-19], :months)
2
iex(2)> Timex.diff(~D[2018-01-20], ~D[2017-10-19], :months)
4
```

### Description of issue

The expected behaviour would be:

```elixir
iex(1)> Timex.diff(~D[2017-12-20], ~D[2017-10-19], :months)
2
iex(2)> Timex.diff(~D[2018-01-20], ~D[2017-10-19], :months)
3
```

We suspect the problem is in this private function: https://github.com/bitwalker/timex/blob/master/lib/comparable/diff.ex#L115-L127

More specifically this logic: https://github.com/bitwalker/timex/blob/master/lib/comparable/diff.ex#L118-L125

The logic should be something like:

```elixir
if d2 >= d1, do: 12 - (m1 - m2), else: 12 - (m1 - (m2 - 1))
```

I'm working on a PR for it!

kelvinst:
Fixes #382


lib/comparable/diff.ex
```diff
@@ -106,34 +106,20 @@ defmodule Timex.Comparable.Diff do
...
     end
   end
-  defp do_diff_months({y,m,_}, {y,m,_}), do: 0
+  defp do_diff_months({y, m, _}, {y, m, _}), do: 0
   defp do_diff_months({y1, m1, d1}, {y2, m2, d2}) when y1 <= y2 and m1 < m2 do
     year_diff = y2 - y1
-    month_diff = if d2 >= d1, do: m2 - m1, else: (m2-1)-m1
-    (year_diff*12)+month_diff
+    month_diff = if d2 >= d1, do: m2 - m1, else: (m2 - 1) - m1
+    (year_diff * 12) + month_diff
```
kelvinst: I did some style adjustments also, hope you don't mind...

lib/comparable/diff.ex
```diff
@@ -106,34 +106,20 @@ defmodule Timex.Comparable.Diff do
...
-          12 - (m1-m2)
-        d2 > d1 ->
-          12 - ((m1-1)-m2)
-        d2 < d1 ->
-          12 - (m1-m2)
-      end
-    (year_diff*12)+month_diff
+  defp do_diff_months({y1, m1, d1}, {y2, m2, d2}) when y1 < y2 and m1 > m2 do
+    year_diff = y2 - (y1 + 1)
+    month_diff = if d2 >= d1, do: 12 - (m1 - m2), else: 12 - (m1 - (m2 - 1))
```
kelvinst: This is the logic that changed... It was completely wrong

lib/comparable/diff.ex
```diff
@@ -106,34 +106,20 @@ defmodule Timex.Comparable.Diff do
...
-      cond do
-        d2 > d1 ->
-          11
-        :else ->
-          12
-      end
-    (year_diff*12)+month_diff
+  defp do_diff_months({y1, m, d1}, {y2, m, d2}) when y1 < y2 do
+    year_diff = y2 - (y1 + 1)
+    month_diff = if d2 >= d1, do: 12, else: 11
```
kelvinst: This is the other logic that changed... Also wrong...

test/timex_test.exs
```diff
@@ -363,6 +363,26 @@ defmodule TimexTests do
...
+    assert Timex.diff(~D[2018-09-19], ~D[2017-10-19], :months) == 11
+    assert Timex.diff(~D[2018-09-20], ~D[2017-10-19], :months) == 11
+
+    assert Timex.diff(~D[2018-10-18], ~D[2017-10-19], :months) == 11
+    assert Timex.diff(~D[2018-10-19], ~D[2017-10-19], :months) == 12
+    assert Timex.diff(~D[2018-10-20], ~D[2017-10-19], :months) == 12
+
+    assert Timex.diff(~D[2018-11-18], ~D[2017-10-19], :months) == 12
+    assert Timex.diff(~D[2018-11-19], ~D[2017-10-19], :months) == 13
+    assert Timex.diff(~D[2018-11-20], ~D[2017-10-19], :months) == 13
```
kelvinst: Hope you're OK with this test format...

lib/comparable/diff.ex
tubedude: @kelvinst 

I like the following logic:

```
(year(d1) * 12 + month(d1)) - (year(d2) * 12 + month(d2))
```

It does not use if logic, which I think is a positive thing.

kelvinst: Well, I do prefer it too, but this way we would not be considering the day of the month on the diff, and I particularly **need** to consider the day on my use case.

I'll try to think how we can still consider the day and get a better code on this. Thanks for the suggestion.

kelvinst:
@tubedude what about it now? I've changed the logic completely to:

- Converts the two dates to a float representing the months following the logic:
  - `(year * 12) + month + (day / 32)`, were we divide the day by 32 to always give us a value less than 0 that represents the day of the month
- Then just subtracts the two given floats
- And finally does round the value following the rule:
  - When it is a positive value, `Float.floor/1`'s it
  - Otherwise, `Float.ceil/1`'s it

This way, all the test scenarios passed, which gives me confidence we're backward compatible!

bitwalker:
Can you test against master (i.e. without your changes)? I've recently merged some long standing PRs for diffing/shifting which I think may address issues you had. If not, we can revisit this PR then.

ckhrysze:
@kelvinst I ran the tests in this PR against the current master, and they all pass with the other changes bitwalker made or merged in. If there are still things you'd like to see from your efforts, we can definitely have a conversation around that. Otherwise I'll close this PR fairly soon.

kelvinst:
Yep, @bitwalker solution on master worked well and the tests I added are passing! I updated this branch and you can merge it if you want more examples on the tests.

Issue #352, opened 2017-08-04
Microsecond precision is lost on shift operations


alvinlindstam:
### Steps to reproduce

```
iex(228)> n = Timex.now()                                                  
#<DateTime(2017-08-04T16:44:51.568241Z Etc/UTC)>
iex(229)> {microseconds, _precision } = n.microsecond                      
{568241, 6}
iex(230)> Timex.shift(n, microseconds: -microseconds).microsecond          
{0, 0}
iex(231)> Timex.shift(n, microseconds: -rem(microseconds, 100)).microsecond
{568200, 4}
```
### Description of issue
When a datetime is shifted, the microseconds seem to reduce the precision to whatever could be assumed based on the number of trailing zeroes.

I would expect all of the datetimes above to keep precision at 6.

I don't really know what I'd expect if the precision is low on the input DateTime, and single digit microseconds are shifted.

bitwalker:
Yeah we definitely should not be reducing precision when we were given an input with a precision already set.

novaugust:
This bug is still around. The PR to fix it actually only fixed the implementation for `%NaiveDateTime{}`

Here's a repro on 3.3
```
iex(2)> n = Timex.now()
#DateTime<2019-04-22 15:55:59.005002Z>
iex(3)> {microseconds, _precision } = n.microsecond
{5002, 6}
iex(4)> Timex.shift(n, microseconds: -microseconds).microsecond
{0, 0}
iex(5)> Timex.shift(n, microseconds: -rem(microseconds, 100)).microsecond
{5000, 3}
```

novaugust:
Buuut it's fixed on 3.5! So nevermind and thanks =)

PR #386, opened 2017-10-25
Fix problems with shift implementation and add tests

fsabe:
This was motivated by issue #352, but I found a few more problems.

- Fix microseconds precision being lost;
- Fix error when shifting to an invalid value by microseconds;
- Unify clauses to remove code duplication.

As @alvinlindstam mentioned, I'm also not sure how to handle precision when shifting by a value with higher precision.


bitwalker:
This looks great! Nice work!

Issue #388, opened 2017-10-27
Fractional seconds are not padded with zeroes when formatting


fsabe:
```elixir
iex> t                            
#<DateTime(2014-11-03T01:41:02.023Z Etc/UTC)>
iex> Timex.format(t, "%f", :strftime)
{:ok, "230"}
iex> Timex.format(t, "{ss}")
{:ok, ".230"}
```

I expected the result to be `{:ok, "023000"}` and `{:ok, ".023"}`.

fsabe:
This affects all formatters that include fractional seconds in the output.

e.g.
```elixir
iex> Timex.format(t, "{RFC3339}")
{:ok, "2014-11-03T01:41:02.230+00:00"}
```

bitwalker:
I'm not clear on why `%f` should return `"023000"`, but you are correct in that the fractional seconds appear to be padded incorrectly.

bitwalker:
Ah I see, nevermind, moving discussion to #389

PR #389, opened 2017-10-27
Fix microseconds formatting

fsabe:
This fix formatting fractional seconds with leading zeroes (issue #388) and change the format of microseconds when using `%f` to conform with what's documented:

>`%f` - microseconds in zero padded decimal form, i.e. 025000

Please see the changed tests. I'd like to hear some thoughts on whether the precision should be taken into account when formatting microseconds.

### Summary of changes

- Pad with zeroes before truncating when formating fractional seconds;
- Pad with zeroes and never truncate when formatting with %f.


bitwalker:
Thanks for the fix!

PR #400, opened 2017-12-05
Add parsing support for languages in the included locales

rozap:
* I'd like to be able to use this library to parse
  dates that are in non-english languages. I could
  see how to translate dates, but not parse them
  from strings.
* This PR enables the following use case

  ```elixir
    require Timex.Translator
    alias Timex.Translator

    Translator.with_locale("fr") do
      Timex.parse("décembre", "{Mfull}")
    end
    # {:ok, ~N[0000-12-01 00:00:00]}
  ```

Not sure if this is a reasonable strategy, but
there are several obvious implications of this:

Questions:
  * For users with a non-english default locale, this
    will break things
    * Is the global gettext locale the right place to do this?
    * If not, we could explicitly pass the locale in to the parse
      function, which might be less surprising for folks

### Summary of changes

### Checklist

### DIFF
--- a/lib/parse/datetime/helpers.ex
+++ b/lib/parse/datetime/helpers.ex
@@ -3,46 +3,94 @@ defmodule Timex.Parse.DateTime.Helpers do
   import Combine.Parsers.Base
   import Combine.Parsers.Text, except: [integer: 0, integer: 1]
   alias Combine.Parsers.Text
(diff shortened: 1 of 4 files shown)


rozap:
I've been fiddling with this failing test case in `test/nonenglish_parser_test.exs` and I'm not sure what the deal is.

PR #403, opened 2017-12-07
Fixes #402

### LINKED ISSUE #402: format sometimes crashes with a bad match, on some timezones
In two timezones that I've found, the `format/2` function will crash while formatting the offset token. 
### Steps to reproduce
```elixir
iex(44)> dt = %DateTime{                                   
...(44)>   calendar: Calendar.ISO,
...(44)>   day: 9, hour: 6, microsecond: {939190, 6},
...(44)>   minute: 39, month: 9, second: 0,
...(44)>   std_offset: 0, time_zone: "Etc/UTC",
...(44)>   utc_offset: 0, year: 2017, zone_abbr: "UTC"
...(44)> } |> Timex.Timezone.convert("Canada/Newfoundland")
#DateTime<2017-09-09 04:09:00.939190-02:30 NDT Canada/Newfoundland>
iex(46)> Timex.format(dt, "{ISO:Extended}")
{:error, {:badmatch, ["-", "02-30"]}}
```

### Description of issue
Other times seem to format just fine. I found this via property tests in my project that would fail with the bad match on these two timezones:
```
Pacific/Marquesas
Canada/Newfoundland
```

I'll submit a PR asap.

rozap:
* Fix bug where some offsets were expressed with a dash
  between hours and minutes, and the format/2 function would
  crash while formatting the offset

### Summary of changes

### Checklist

### DIFF
--- a/lib/format/datetime/formatter.ex
+++ b/lib/format/datetime/formatter.ex
@@ -651,8 +651,12 @@ defmodule Timex.Format.DateTime.Formatter do
       {:error, _} = err -> err
       "" -> ""
       offset ->
-        [qualifier, <<hour::binary-size(2), min::binary-size(2)>>] = String.split(offset, "", [trim: true, parts: 2])
-        <<qualifier::binary, hour::binary, ?:, min::binary>>
+        case String.split(offset, "", [trim: true, parts: 2]) do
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks!

PR #408, opened 2018-01-29
Add documentation concerning interval test

edouardmenayde:
### Summary of changes

I added documentation for interval testing because I was searching for the feature in the doc found nothing and finally found it in the code thus I want to improve the documentation.

I could not document directly the implementation of `Enumerable` and found no way of doing it.


bitwalker:
Thanks!

PR #412, opened 2018-02-13
Timex.format when invalid date passed

Apelsinka223:
### Summary of changes

When wrong date passed to Timex.format!/1 with `"{M}"` format (maybe some others too), it raises
    `` ** (UndefinedFunctionError) function :error.month/1 is undefined (module :error is not available)``
I changed it to ``** (Timex.TimexError) invalid_date``

### Checklist


lib/format/datetime/formatter.ex
```diff
@@ -40,22 +40,29 @@ defmodule Timex.Format.DateTime.Formatter do
...
-    do
-      date = case date do
-         %{__struct__: struct} when struct in [Date, DateTime, NaiveDateTime, Time] -> date
-         _other -> Timex.to_naive_datetime(date)
-      end
-      case formatter.lformat(date, format_string, locale)do
-        {:ok, result}    -> result
-        {:error, reason} -> raise FormatError, message: reason
+    when is_binary(format_string) and is_binary(locale) and is_atom(formatter) do
+    case Timex.to_naive_datetime(date) do
```
bitwalker: We don't want to convert to NaiveDateTime in general here, only for things passed in which are not one of the built-in Elixir calendar types. Was there a specific reason you made this change?

lib/format/datetime/formatter.ex
```diff
@@ -80,7 +87,7 @@ defmodule Timex.Format.DateTime.Formatter do
       try do
         {:ok, lformat!(date, format_string, locale, formatter)}
       catch
-        _type, %FormatError{:message => msg} ->
```
bitwalker: I think you need to handle both errors here

lib/timex_error.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Timex.TimexError do
+  @moduledoc """
+  Used for errors encountered during string formatting.
```
bitwalker: This is what `FormatError` is for, I think you can remove this and use `ArgumentError` where you've used `TimexError` in this PR, as that is more appropriate anyway. If you don't agree, we can discuss it further though

bitwalker:
Thanks @Apelsinka223! I have some changes I've requested to this, but it looks like there was for sure a bug in there (returning an error tuple from `lformat!` rather than raising in that one case). The other changes don't appear to be related though, could you explain them further?

lib/format/datetime/formatter.ex
Apelsinka223: I separeted method with matching of date struct (line 49 and line 57) just for readability. I added in logic only check 
 of `Timex.to_naive_datetime(date)` result

lib/format/datetime/formatter.ex
Apelsinka223: `%FormatError{:message => msg}` will match with `%{:message => msg}`, since processing of them is same I decided to remove this case

lib/timex_error.ex
Apelsinka223: You are right, ArgumentError will be more suitable, I'll fix this

Apelsinka223:
Actual fix is in this case
https://github.com/bitwalker/timex/pull/412/files#diff-f9c18838de3da12d2fc025790c7ab1a3R59
I added raise in lformat!, but it not fix of this problem.

lib/format/datetime/formatter.ex
bitwalker: Sure I see now, that's fine 👍

bitwalker:
Thanks!

Issue #414, opened 2018-02-27
Timex.format spec does not take into account Time.t


jeremyowensboggs:
### Steps to reproduce

`Timex.format!(~T[12:00:00], "{h12}:{m} {AM}")` will produce the expected result of `"12:00 PM"`
However dialyzer flags it as an error because Time.t is not one of the allowed arguments.

PR #415, opened 2018-02-27
Add Time.t to the calendar_types and valid_datetime

jeremyowensboggs:
### Summary of changes
In regards to https://github.com/bitwalker/timex/issues/414
Timex.format!(~T[12:00:00], "{h12}:{m} {AM}") will produce the expected result of "12:00 PM"
However dialyzer flags it as an error because Time.t is not one of the allowed arguments.

### Checklist

- [ x ] New functions have typespecs, changed functions were updated
- [ x ] Same for documentation, including moduledocs
- [ x ] Tests were added or updated to cover changes
- [ x ] Commits were squashed into a single coherent commit


bitwalker:
Thanks!

PR #421, opened 2018-03-16
Fix infinite loop on Timex.Timezone.convert/2

enerick:
### Summary of changes

This PR fixes an infinite loop on Timex.Timezone.convert/2.

I found the use of `Timex.Timezone.convert(nil, "Etc/UTC")` leads non-termination due to the lack of response handling for Timex.to_datetime/1 in the last clause of Timex.Timezone.convert/2.

### Checklist

- ~~[ ] New functions have typespecs, changed functions were updated~~
- ~~[ ] Same for documentation, including moduledocs~~

### DIFF
--- a/lib/timezone/timezone.ex
+++ b/lib/timezone/timezone.ex
@@ -472,7 +472,10 @@ defmodule Timex.Timezone do
     end
   end
   def convert(date, tz) do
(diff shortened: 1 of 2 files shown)


bitwalker:
Good catch, thanks!

PR #422, opened 2018-03-20
Fix Timex.Parse.DateTime.Tokenizer.tokenize specs

barthez:
### Summary of changes

Fixed Timex.Parse.DateTime.Tokenizer.tokenize callback definition and specs in all modules that implements its behaviour. It was causing dialyzer to fail in my project.

### Checklist


bitwalker:
Thanks!

Issue #398, opened 2017-11-28
strftime formatting of milliseconds


jurta:
### Steps to reproduce

- `Timex.now |> Timex.format!("%Y/%m/%d %T.%L", :strftime)`
- or just `Timex.now |> Timex.format!("%L", :strftime)`
- fails with:
```
** (Timex.Format.FormatError) got {:formatter, :unsupported_token, :ms} while retrieving Exception.message/1 for %Timex.Format.FormatError{message: {:formatter, :unsupported_token, :ms}} (expected a string)
    (timex) lib/format/datetime/formatter.ex:58: Timex.Format.DateTime.Formatter.lformat!/4
```

### Description of issue

- Expected to return a string with milliseconds (000..999) as described in the documentation.

Thanks!

prook:
The FormatError reason handling may be improved, but the original problem remains: "%L" blows up with `{:formatter, :unsupported_token, :ms}`, because Timex.Format.DateTime.Formatter lacks format_token for `:ms` type.

PR #433, opened 2018-06-05
#398 - Add support for millisecond formatting.

prook:
This is my rudimentary attempt at fixing #398.

I'm not sure the min/max width computation is really needed, as the width is forced to 3 by definition, but mimicking microseconds, I put it there.


bitwalker:
Closing in favor of #455

PR #439, opened 2018-07-12
Clarify Interval docs about left_open and right_open

nathanl:
### Summary of changes

Adds a couple sentences to the Interval docs to clarify the meaning of `left_open` and `right_open`.


bitwalker:
@nathanl Thanks Nathan! I ended up reworking them in more detail, so I'm going to close this, but thanks for opening a PR!

PR #440, opened 2018-07-12
Fix Manual offsets involving minutes

michaelkschmidt:
This allows conversions to work that specify raw offsets

```elixir
DateTime.utc_now() |> Timex.Timezone.convert("+05:30")
```

parse_offset() returns a suffix which later gets passed back to parse_offset() a second time.  Thus the return has to match what was passed in


bitwalker:
This works in master now, so I'm going to close this. Thanks for the contribution!

PR #441, opened 2018-07-13
Add Interval.contains?/2 to test if one interval contains another

nathanl:
### Summary of changes

Add `Interval.contains?/2` to test if one interval contains another.

This is useful when, for example, deciding if a proposed appointment time slot fits inside a window of availability.

```elixir
Timex.Interval.contains?(availability_window, proposed_appointment)
```

### Checklist

I also formatted the new code & tests with `mix format`.


bitwalker:
Nice! Thanks!

Issue #446, opened 2018-07-25
:strftime_iso_kitchen directive (%r) discards previously applied date directives.


sparta-developers:
### Steps to reproduce

```
"July 24, 2018 5:34:04 PM PDT"
  |> Timex.parse!("%B %e, %Y %r %Z", :strftime)
  |> Timex.format!("%B %e, %Y %r %Z", :strftime)
```

The output of this is `"July 25, 2018 05:34:04 PM PST8PDT"`

### Description of issue

I would expect the conversion to remain July 24.

sparta-developers:
We discovered this bug on version `3.3.0`, but reproduced it on master.

ckhrysze:
The fix was merged and included in the 3.4.1 release, thank you!

PR #447, opened 2018-07-25
Do not discard previously parsed date when applying %r directive

sparta-developers:
### Summary of changes

This fixes issue #446, where parsing using `:strftime` with `%r` overwrites previously parsed date information with the current date. So, when parsing the date `July 21, 2017 05:15:45 PM PST` with `%B %d, %Y %r %Z`, the time is correct but the date is set to today.

This fix does introduce a *different* edge case, where if you only include part of the date, the non-included fields will not be set. So `July 21, 05:15:45 PM` will return a date without a year. That seems like more reasonable behavior than setting the date to today, however.

### Checklist


bitwalker:
I agree with your approach here, thanks!

PR #450, opened 2018-07-30
Make beginning_of_month return the first of the month in the correct timezone

mtrudel:
### Summary of changes

Takes the time_offset of the beginning of the month into account when computing `beginning_of_month`. Fixes an issue whereby `beginning_of_month` called with e.g. `13:45 15 Nov 2018 EST` would return `00:00 1 Nov 2018 EST` where it should be `00:00 1 Nov 2018 EDT` since the Eastern timezone was still in daylight savings on 1 Nov. 

Note that other `{beginning,end}_of_{day,week,month}` functions already do the right thing - the implementation of this method was the outlier. 

### Checklist


bitwalker:
Thanks!

Issue #454, opened 2018-08-08
Timex.Format.FormatError - invalid format: { :formatter, :unsupported_token, :ms }


simpers:
### Steps to reproduce
Currently using Timex version 3.4.0 through github directly using `{ :timex, github: "bitwalker/timex", tag: "3.4.0" }`.

I was writing a unit test for our time formatting when I stumbled upon this. In my actual case `format_string` is a constant in the app itself, and not defined directly in the unit test. But other than that it is the same code as I have. The directive `%f` works, but is not what we need.

- API function called
- Arguments provided

```
test "Test formatting a timestamp with %L", _state do
    format_string = "%FT%T.%LZ"
    time = "2018-01-01T00:00:00.000Z"
    |> Timex.parse!(format_string, :strftime)
    |> Timex.format!(format_string, :strftime)
    assert time == "2018-01-01T00:00:00.000Z"
end
```

### Description of issue

- What are the expected results?
As per the documentation [describing the Strftime formatter](https://hexdocs.pm/timex/Timex.Format.DateTime.Formatters.Strftime.html#content) there should be a directive, `%L`, that produces: `milliseconds (000..999)`

- If you are certain this is a bug in Timex, do you have references I
  can use to check the algorithm/logic against? For instance, a well-established
  date/time library in another language, a paper, test suite, etc. It is by no
  means a requirement, but is immensely helpful in fixing bugs, and provides me
  extra data to use when building out the test suite

I'm sure it is a bug as there is documentation on it (perhaps by mistake, but then the documentation has a bug, I suppose). It is also quite a common way of formatting a timestamp so I would expect it to exist, as it does in other libraries (not sure of which one, but on the Java side gson formats it this way currently in our environment, and I am trying to match it up to this).

simpers:
Noticed that there was a pull request #433 regarding this issue already. Didn't find it initially. So I guess this is a duplicate.

PR #455, opened 2018-08-08
#454: Fixed this and added some tests.

simpers:
Fixed to 3 zeroes as that is t…he point of the %L, I would assume. If that assumption incorrect, feedback is welcome.

### Summary of changes

The directive `%L` causes a crash.

### Checklist

The tests were mostly copied from #433 and adapted, and made some more. Not sure whether the division should be rounded, but I would assume so. It might not be the case, and if so that can be fixed by just doing `div(us, 1000)` similarly to what #433 did.

### DIFF
--- a/lib/format/datetime/formatter.ex
+++ b/lib/format/datetime/formatter.ex
@@ -598,8 +598,10 @@ defmodule Timex.Format.DateTime.Formatter do
   def format_token(_locale, :us, _date, _modifiers, flags, width) do
     pad_numeric(0, flags, width)
   end
-  def format_token(_locale, :ms, _date = %{microsecond: {us, _precision}}, _modifiers, flags, width), do: pad_numeric(Kernel.round(us / 1000), flags, width_spec(3..3))
-  def format_token(_locale, :ms, _date, _modifiers, flags, width), do: pad_numeric(0, flags, width)
+  def format_token(_locale, :ms, _date = %{microsecond: {us, _precision}}, _modifiers, flags, width),
+    do: pad_numeric(Kernel.round(us / 1000), flags, width_spec(3..3))
+  def format_token(_locale, :ms, _date, _modifiers, flags, width),
+    do: pad_numeric(0, flags, width)
   def format_token(locale, :am, %{hour: hour}, _modifiers, _flags, _width) do
     day_periods = Translator.get_day_periods(locale)
     {_, am_pm} = Timex.Time.to_12hour_clock(hour)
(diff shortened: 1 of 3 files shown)


lib/format/datetime/formatter.ex
```diff
@@ -598,6 +598,8 @@ defmodule Timex.Format.DateTime.Formatter do
   def format_token(_locale, :us, _date, _modifiers, flags, width) do
     pad_numeric(0, flags, width)
   end
+  def format_token(_locale, :ms, _date = %{microsecond: {us, _precision}}, _modifiers, flags, width), do: pad_numeric(Kernel.round(us / 1000), flags, width_spec(3..3))
```
bitwalker: Can you split this on two lines, by adding a newline after the `,` and before `do:`? Same with the next clause

test/format_strftime_test.exs
```diff
@@ -17,6 +17,20 @@ defmodule DateFormatTest.FormatStrftime do
     end)
   end
 
+  test "success!" do
+    date = Timex.to_datetime({2013,03,02})
+
+    formatter = Timex.Format.DateTime.Formatters.Strftime
```
bitwalker: You don't need this, just use the `format/2` local which is defined in this module, like the other tests

test/format_strftime_test.exs
```diff
@@ -17,6 +17,20 @@ defmodule DateFormatTest.FormatStrftime do
...
+    assert "2013-03-02" == formatter.format!(date, "%Y-%m-%d")
+  end
+
+  test "error tuple" do
+    date = Timex.to_datetime({2013,03,02})
+
+    formatter = Timex.Format.DateTime.Formatters.Strftime
```
bitwalker: Same as the above

test/format_strftime_test.exs
bitwalker: Actually why is this test added? I think it can be removed entirely, but I'm open to being swayed if you think it is necessary

test/format_strftime_test.exs
bitwalker: Let's give this test a more descriptive name, like "error returned when invalid directive is given"

test/format_strftime_test.exs
simpers: I just ran the coverage report and saw that this module was not hitting all lines, so I thought I'd just add two tests to cover it completely. So you're right in that it does not affect what I added, but I thought that I should raise the coverage instead of decrease it :)

simpers:
Just did a rebase on the newest master and pushed again. I am assuming that the checks failed due to the new master and not something I did, as I just took that feedback and fixed some details.

PR #458, opened 2018-08-16
inclusive_start/inclusive_end options for Timex.between?

glassbead0:
* Timex.between? the :inclusive option can now take `:start`, `:end` in addition to `true`
### Summary of changes

I have a use case to ask if a given time is between two other times, with one bound being inclusive, and the other being exclusive. I imagine this is useful for anyone iterating over adjacent intervals of time.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -770,27 +770,38 @@ defmodule Timex do
   Returns a boolean indicating whether the first `Timex.Comparable` occurs between the second
   and third.
 
-  By default, the `start`and `ending` bounds are *exclusive*. You can opt for inclusive bounds
-  by setting the `:inclusive` option to `true`.
-
+  By default, the `start` and `end` bounds are *exclusive*. You can opt for inclusive bounds with the
+  `inclusive: true` option.
+  To set just one of the bounds as inclusive, use the
+  `inclusive: :start` or `inclusive: :end` option.
   """
   @type between_options :: [
     inclusive: boolean
+    | :start
+    | :end
   ]
   @spec between?(Time, Time, Time, between_options) :: boolean | {:error, term}
   @spec between?(Comparable.comparable, Comparable.comparable, Comparable.comparable, between_options) ::
     boolean | {:error, term}
   def between?(a, start, ending, options \\ []) do
-    inclusive = Keyword.get(options, :inclusive, false)
+    [inclusive_start, inclusive_end] =
+    case Keyword.get(options, :inclusive, false) do
+      :start -> [true, false]
+      :end -> [false, true]
+      true -> [true, true]
+      _ -> [false, false]
+    end
 
     after_start = compare(a, start)
     before_ending = compare(a, ending)
 
-    case {inclusive, after_start, before_ending} do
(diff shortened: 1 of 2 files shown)


lib/timex.ex
```diff
@@ -771,26 +771,33 @@ defmodule Timex do
...
-  by setting the `:inclusive` option to `true`.
+  by setting the `:inclusive` option to `true`. You can also specify `:inclusive_start` and/or
+  `:inclusive_end` options, which will override the `:inclusive` option for the given bound.
 
   """
   @type between_options :: [
-    inclusive: boolean
+    inclusive: boolean,
+    inclusive_start: boolean,
+    inclusive_end: boolean
```
pezra: it might be nicer to expand the domain of `inclusive` rather than introducing new options. for example:

```
iex> Timex.between?(t1, t2, t3, inclusive: true)
iex> Timex.between?(t1, t2, t3, inclusive: :start)
iex> Timex.between?(t1, t2, t3, inclusive: :end)
```
that would avoid some awkward potential situations where `inclusive` and `inclusive_start` and `inclusive_end` disagree with one another.

```
iex> Timex.between?(t1, t2, t3, inclusive: false, inclusive_start: true)
iex> Timex.between?(t1, t2, t3, inclusive: true, inclusive_start: false)
```

lib/timex.ex
```diff
@@ -770,27 +770,39 @@ defmodule Timex do
   Returns a boolean indicating whether the first `Timex.Comparable` occurs between the second
   and third.
 
-  By default, the `start`and `ending` bounds are *exclusive*. You can opt for inclusive bounds
-  by setting the `:inclusive` option to `true`.
+  By default, the `start`and `ending` bounds are *exclusive*. You can opt for inclusive `start`
+  and `end` bounds by setting the `:inclusive` option to `true`. You can also specify
+  `:inclusive_start` and/or `:inclusive_end` options, which will override the `:inclusive`
+  option for the given bound.
```
ckhrysze: I like the updated option! My only ask here is that the doc reflect the latest code update and consistent use of either 'end' or 'ending'.

glassbead0: 👍 Thanks for the reminder. I've updated the doc.

lib/timex.ex
```diff
@@ -770,27 +770,38 @@ defmodule Timex do
...
     inclusive: boolean
+    | :start
+    | :end
   ]
   @spec between?(Time, Time, Time, between_options) :: boolean | {:error, term}
   @spec between?(Comparable.comparable, Comparable.comparable, Comparable.comparable, between_options) ::
     boolean | {:error, term}
   def between?(a, start, ending, options \\ []) do
-    inclusive = Keyword.get(options, :inclusive, false)
+    [inclusive_start, inclusive_end] =
```
bitwalker: @ckhrysze Can we make this a tuple rather than a list? I missed the review on this, sorry. As a rule I prefer tuples for a known-size set of things, with small N. It better reflects the purpose, access is constant, and is a smaller in-memory structure (though obviously for cases like this, that isn't really important). You can commit the change directly to master, no need to go through another PR.

Issue #426, opened 2018-04-07
Recommended way to add ordinal suffix for the day of month?


parlays:
I cannot find the ordinal suffix for the month.  Similar to what PHP has for their datetime:
English ordinal suffix for the day of the month, 2 characters | st, nd, rd or th.
http://php.net/manual/en/function.date.php

Is a custom formatter recommended to add the suffix?

parlays:
For now I created this module to get the suffix and then I put the result in the Timex.format call.

defmodule OrdinalSuffix do

  def get(i) do
    value = elem(Integer.parse(i), 0)
    j = div(value, 10)
    k = div(value, 100)

    cond do
      j == 1 && k != 11 ->
        "st"
      j == 2 && k != 12 ->
        "nd"
      j == 3 && k != 13 ->
        "rd"
      true ->
        "th"
    end

  end
  
end

bitwalker:
I'm open to extending the formatter with a new token, if you would like to submit a PR!

PR #463, opened 2018-08-29
Added new directive: day with suffix

yangel:
### Summary of changes

Addressed issue described in #426 

### Checklist

### DIFF
--- a/lib/format/datetime/formatter.ex
+++ b/lib/format/datetime/formatter.ex
@@ -476,6 +476,17 @@ defmodule Timex.Format.DateTime.Formatter do
     do: pad_numeric(date.day, flags, width)
   def format_token(_locale, :oday, date, _modifiers, flags, width),
     do: pad_numeric(Timex.day(date), flags, width)
+  def format_token(_locale, :dsuff, %{:day => day}, _modifiers, _flags, _width) do
+    remainder = day |> rem(10)
+    suffix =
+      cond do
+        remainder == 1 && day != 11 -> "st"
+        remainder == 2 && day != 12 -> "nd"
+        remainder == 3 && day != 13 -> "rd"
+        true -> "th"
+      end
+    "#{day}" <> suffix
+  end
   # Weeks
   def format_token(_locale, :iso_weeknum, date, _modifiers, flags, width) do
     {_, week} = Timex.iso_week(date)
--- a/lib/format/datetime/formatters/strftime.ex
+++ b/lib/format/datetime/formatters/strftime.ex
@@ -22,8 +22,8 @@ defmodule Timex.Format.DateTime.Formatters.Strftime do
   * `-`       - don't pad numerical results (overrides default padding if any)
   * `0`       - use zeros for padding
   * `_`       - use spaces for padding
-  * `:`, `::` - used only in combination with `%z`; see description of `%:z`
-                and `%::z` below
+  * `:`, `::` - used only in combination with `%z` and `%d`; see description of `%:d`,
+                `%:z` and `%::z` below
 
   `<width>` is a non-negative decimal number specifying the minimum field
   width.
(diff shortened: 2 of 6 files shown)


bitwalker:
These type of numbers are called ordinals, so I prefer to use that terminology when possible. Since ordinals can come with a suffix or without in written form, this can still be somewhat confusing, but from a maintenance point of view, I'd like to clarify things in here by saying "ordinal suffix" rather than "day suffix", since the latter isn't clear to me what it means on first read.

Rather than add a new directive for this though, I think it may make more sense to add it as a flag (i.e. if the current directive is represented by an ordinal number, require the suffix in printed form - if the current directive is not an ordinal number, then the flag is invalid, and either ignored or results in an error). This means we can easily add it to both the default and strftime formats by introducing a corresponding flag, and would support arbitrary ordinals with suffixes (e.g. if you wanted a format that says something like "the 10th hour of the 3rd day of March" or something like that).

I haven't yet thought through what the flag would be in the two formats, but let's discuss the approach first, and we can hash that out.

zorn:
I tried to use the code in this PR and noticed issues with some numbers. I eventually rewrote it like this:

```
  def ordinal_suffix(value) when value in [1, 21, 31] do
    "st"
  end

  def ordinal_suffix(value) when value in [2, 22] do
    "nd"
  end

  def ordinal_suffix(value) when value in [3, 23] do
    "rd"
  end

  def ordinal_suffix(value) do
    "th"
  end
```

With tests:

```
  describe "ordinal_suffix/1" do
    test "valid data scenarios" do
      assert EventDateFormatter.ordinal_suffix(1) == "st"
      assert EventDateFormatter.ordinal_suffix(2) == "nd"
      assert EventDateFormatter.ordinal_suffix(3) == "rd"
      assert EventDateFormatter.ordinal_suffix(4) == "th"
      assert EventDateFormatter.ordinal_suffix(5) == "th"
      assert EventDateFormatter.ordinal_suffix(6) == "th"
      assert EventDateFormatter.ordinal_suffix(7) == "th"
      assert EventDateFormatter.ordinal_suffix(8) == "th"
      assert EventDateFormatter.ordinal_suffix(9) == "th"
      assert EventDateFormatter.ordinal_suffix(10) == "th"
      assert EventDateFormatter.ordinal_suffix(11) == "th"
      assert EventDateFormatter.ordinal_suffix(12) == "th"
      assert EventDateFormatter.ordinal_suffix(13) == "th"
      assert EventDateFormatter.ordinal_suffix(14) == "th"
      assert EventDateFormatter.ordinal_suffix(15) == "th"
      assert EventDateFormatter.ordinal_suffix(16) == "th"
      assert EventDateFormatter.ordinal_suffix(17) == "th"
      assert EventDateFormatter.ordinal_suffix(18) == "th"
      assert EventDateFormatter.ordinal_suffix(19) == "th"
      assert EventDateFormatter.ordinal_suffix(20) == "th"
      assert EventDateFormatter.ordinal_suffix(21) == "st"
      assert EventDateFormatter.ordinal_suffix(22) == "nd"
      assert EventDateFormatter.ordinal_suffix(23) == "rd"
      assert EventDateFormatter.ordinal_suffix(24) == "th"
      assert EventDateFormatter.ordinal_suffix(25) == "th"
      assert EventDateFormatter.ordinal_suffix(26) == "th"
      assert EventDateFormatter.ordinal_suffix(27) == "th"
      assert EventDateFormatter.ordinal_suffix(28) == "th"
      assert EventDateFormatter.ordinal_suffix(29) == "th"
      assert EventDateFormatter.ordinal_suffix(30) == "th"
      assert EventDateFormatter.ordinal_suffix(31) == "st"
    end
  end
```
I am making the assumption there is never more than 31 days.

Thanks for the helpful library.

darraghenright:
Hello there 👋 

What's the latest with this PR? It's a couple of years old now, but if it's still a desired feature I'd be happy to do further work on it.

tfantina:
I just came across this attempting to create a label for some graphs I'm making.  I'd be happy to pick up the work on this rather than, effectively, rewriting it in my codebase.

PR #464, opened 2018-08-31
Added Afrikaans translations

juliandicks:
### Summary of changes

Added afrikaans translations (af)
> because we need Afrikaans translations for a website I'm working on

### Checklist

- [x] Translations checked and confirmed
- [x] Tests pass


ckhrysze:
@juliandicks Thank you for the translation work! An update was made on master to restore 1.4 compatibility. If you update the PR with that change, travis should be happy and we can get this PR merged in.

PR #469, opened 2018-09-28
Added missing type specs for Timex.set/2

thomasbrus:
### Summary of changes

In https://github.com/bitwalker/timex/pull/376, the option was added to pass a `%Time{}` struct to Timex.set/2 via `:time`. However currently dialyzer complains because the type is still specified as {hour, minute, second}. This PR fixes that:)

### Checklist


thomasbrus:
@ckhrysze agreed, hadn't noticed! See last commit. Also I moved the tests for set/2 into a new file. Let me know if you have any other comments

PR #477, opened 2018-10-12
WIP beginning_of_seconds and end_of_seconds

nathanl:
These are two functions I found useful in my work, where I needed to "round" some datetimes to the nearest hour. They seems to fit in with functions like `end_of_month`, etc.

This is WIP, as it needs to be split up into protocol implementations. But I wanted to see if you like the idea before doing that.

### Summary of changes

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -1362,6 +1362,65 @@ defmodule Timex do
 
   def end_of_quarter(_, _), do: {:error, :invalid_year_or_month}
 
+
+  @doc """
+  Given a Datetime, returns a datetime at the end of a period of N seconds.
+  For example, this could be used to round to the next hour.
+
+  iex> Timex.end_of_seconds(~N[2018-01-01 00:00:13], 15)
+  ~N[2018-01-01 00:00:15]
+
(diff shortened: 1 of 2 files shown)


lib/timex.ex
```diff
@@ -1362,6 +1362,66 @@ defmodule Timex do
...
+    if rem(unix, seconds) == 0 do
+      input_datetime
+    else
+      rounded_down = div(unix, seconds) * seconds
+
+      output_datetime = rounded_down |> Timex.from_unix()
+
+      # return the same type we were given
+      case input_datetime do
+        %DateTime{} -> output_datetime
```
nathanl: This would be split into the various protocol implementations

bitwalker:
My feeling is that we'd be better off adding `Timex.round/2` which rounds to the nearest granularity (as seen in other APIs, like `compare/3`). A `Timex.round/3` variant could take an argument which indicates the rounding direction. Alternatively, we could do `Timex.ceil/2` and `Timex.floor/2` which act like their mathematical counterparts, but take a granularity like the proposed `round/2` function. Thoughts?

PR #480, opened 2018-10-30
Setup property based test framework and add sample tests

mkorszun:
#### WHY
Introduce property based tests in order to:
* increase correctness of timex public api by discovering corner cases earlier
* reduce number of example based tests, which will not be able to cover all scenarios and edge cases anyway

#### What
* Use [StreamData](https://github.com/whatyouhide/stream_data) as a property based test framework
* Introduce a couple of sample property tests. We can not really add property tests for all the public functions anyway. It will be a longer process and we would like to start it with this PR.

### Checklist

### DIFF
--- a/test/helpers/property_helpers.ex
+++ b/test/helpers/property_helpers.ex
@@ -1,0 +1,30 @@
+defmodule PropertyHelpers do
+  @moduledoc """
+  Custom data generators to be used in property tests
+  """
+
+  def date_time_generator(:struct) do
+    date_time_generator(:tupple)
+    |> StreamData.map(&Timex.to_datetime/1)
+  end
+
+  def date_time_generator(:tupple) do
+    [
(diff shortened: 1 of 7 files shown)


test/helpers/property_helpers.ex
```diff
@@ -0,0 +1,30 @@
...
+  @moduledoc """
+  Custom data generators to be used in property tests
+  """
+
+  def date_time_generator(:struct) do
+    date_time_generator(:tupple)
+    |> StreamData.map(&Timex.to_datetime/1)
+  end
+
+  def date_time_generator(:tupple) do
```
ckhrysze: I may just be missing something, but why :tupple and not :tuple?

mkorszun: @ckhrysze typo :( - thanks for the feedback 👍

PR #483, opened 2018-11-02
Fixes wrong interval bounds interpretation, fixes #482

### LINKED ISSUE #482: The bounds of intervals are inverted now!
Hi.
In your [commit](https://github.com/bitwalker/timex/commit/063e2dd1d5c0767e5e24ee3b8bc8bd2157aa7362) you swapped the bounds of intervals.
BUT: Your initial implementation was correct, see your link in the docs. Now your code as well as your documentation are just wrong

> An **open interval** does not include its endpoints, and is indicated with parentheses `()`.

> A **closed interval** is an interval which includes all its limit points, and is denoted with square brackets `[]`.

Revert the commit?

Greetings

Andreas

arfl:
Hi.
I said it already in the Issue #482, you misinterpreted the meaning of open/closed bounds of intervals.
Open intervals **do not include** the bounds, closed bounds **do include** the bounds.
So, your original implementation was correct.

### Summary of changes
Since your original implementation was correct, I have reverted your [commit](https://github.com/bitwalker/timex/commit/063e2dd1d5c0767e5e24ee3b8bc8bd2157aa7362) by hand, preserving the documentation changes you did in this commit (just adjusted the wrong statements about the bounds) and the updated implementation.
The 3 broken tests are already in the master branch and are not related to the changes I did.
When you see my diff, you would realize that I changed more lines that I needed. That's because of the formatter I always run. But most of the changes, the formatter has done, are really good ones.

Hope you can fix it as soon as possible. And it would be great if you consider this fix to be worth releasing a new version.

By the way, thank you for this great library! We use it in EVERY project we start. 

### Checklist

- [x ] New functions have typespecs, changed functions were updated
- [x ] Same for documentation, including moduledocs
- [x ] Tests were added or updated to cover changes
- [x ] Commits were squashed into a single coherent commit

### DIFF
--- a/lib/interval/interval.ex
+++ b/lib/interval/interval.ex
@@ -6,11 +6,11 @@ defmodule Timex.Interval do
 
       iex> use Timex
       ...> Interval.new(from: ~D[2016-03-03], until: [days: 3])
-      %#{__MODULE__}{from: ~N[2016-03-03 00:00:00], left_open: true, right_open: false, step: [days: 1], until: ~N[2016-03-06 00:00:00]}
+      %#{__MODULE__}{from: ~N[2016-03-03 00:00:00], left_open: false, right_open: true, step: [days: 1], until: ~N[2016-03-06 00:00:00]}
 
       iex> use Timex
       ...> Interval.new(from: ~D[2016-03-03], until: ~N[2016-03-10 01:23:45])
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks!

PR #487, opened 2018-11-05
Fix a couple of shift issues

baldmountain:
### Summary of changes

- If a shift crosses a daylight savings time shift be sure to include the offset_std difference in the time conversion. This change probably fixes the issue mentioned here: https://github.com/bitwalker/timex/blob/master/lib/datetime/datetime.ex#L249-L251

- Shifting backwards into the previous month had an off by one error.

### Checklist

- [ X] Tests were added or updated to cover changes


schneiderderek:
Is there a timeline for releasing these changes in a new version of the package?

PR #489, opened 2018-12-01
Fix Negative Shifts Across Month Boundaries

michaeljguarino:
There was an off by one error in the implementation of logical shifts
for days.  It effectively caused a shift by `days: -1` on day 1 of a month to
resolve back to the current date.

### Checklist


ckhrysze:
I'm closing this as this exact change plus another was made with PR 487, and released in version 3.5. However, I do want to thank you for the contribution!

PR #493, opened 2018-12-10
add overlap functionality to Timex.Interval

glassbead0:
### Summary of changes

* Find the overlap Interval between two intervals.
* Return {:error, :no_overlap_interval} if intervals
  do not overlap, (including overlapping at one instant)

### Checklist


lib/interval/interval.ex
```diff
@@ -382,6 +382,71 @@ defmodule Timex.Interval do
   def max(%__MODULE__{until: until, right_open: false}), do: until
   def max(%__MODULE__{until: until}), do: Timex.shift(until, microseconds: -1)
 
+  @doc """
+  Returns an Interval representing the intersection between two intervals.
+  If the intervals do not overlap, reutrn {:error, :no_overlap_interval}.
```
heyZeus: misspelling `reutrn`

lib/interval/interval.ex
```diff
@@ -382,6 +382,71 @@ defmodule Timex.Interval do
...
+    cond do
+      Timex.equal?(a.until, b.until)  -> {a.until, determine_bound(a.right_open, b.right_open)}
+      Timex.before?(a.until, b.until) -> {a.until, a.right_open}
+      true                            -> {b.until, b.right_open}
+    end
+  end
+
+  @doc """
+  When calculating overlap, if two intervals share a `from` (or `until`), the overlap
+  interval should have a bound matching the "inner" interval (eg: if either inteval has
```
heyZeus: spelling: `inteval`

lib/interval/interval.ex
```diff
@@ -382,6 +382,71 @@ defmodule Timex.Interval do
   def max(%__MODULE__{until: until, right_open: false}), do: until
   def max(%__MODULE__{until: until}), do: Timex.shift(until, microseconds: -1)
 
+  @doc """
+  Returns an Interval representing the intersection between two intervals.
+  If the intervals do not overlap, return {:error, :no_overlap_interval}.
+  If the intervals overlap at a single instant (regardless of open/closed
+  bounds), also return {:error, :no_overlap_interval}
```
pezra: Perhaps it should return `0` in the case of no overlap rather than an error? Two intervals not overlapping doesn't seem like an error to me.

PR #501, opened 2019-01-15
Leverage dynamic locale with gettext

tarzan:
### Summary of changes

In order to let gettext work with an already implemented gettext of an app (or even accross dependencies), all format functions should use the current locale of gettext instead of the pre-configured default. This makes dynamic language switching possible as well. 

This PR just swaps default for current without mentioning it in the docs. Let me know what you think of this, so I can perfect it and make it release worthy.

### Checklist


bitwalker:
I think this change is a good one, but it _is_ technically a breaking change; do we know whether or not people are relying on the current behavior?

tarzan:
> I think this change is a good one, but it is technically a breaking change; do we know whether or not people are relying on the current behavior?

Yes it is. However, the scenario in which this breaks is, I believe, pretty limited. Considering you would need another backend for Gettext already in your app for which you set the locale to be different from the locale you set as default on Timex. I can't phantom a scenario in which you would actually want that... (even though you still can by setting Gettext locale specifically per backend).

I would suggest writing some documentation about how the behaviour has changed and make it a major version release instead of writing additional configuration options to make it backwards compatible. Is that something you can get behind?

tarzan:
Ping! I still believe this is good to go with minor chances that it breaks for anyone out there. How can I help releasing this?

bitwalker:
We can get this merged and released, but I'd like to have the CHANGELOG updated that mentions that this change is potentially breaking, but that the old behavior was technically a bug - the new behavior fixes that bug and will do the expected thing with regard to localization and integration with gettext.

Once that is done, I'll get this merged and rolled out in the next release

tarzan:
So, I somehow thought you wanted to do the editing of the changelog yourself, I forgot about this open PR in the meantime.

Just noticed it again today and decided to finally write the changelog entry. Also did a full rebase. Hope you're okay with the changes now.

tarzan:
Just wondering if you were able to check the recent changes?

tarzan:
I just rebased on your latest master. How to move forward with this? @bitwalker ?

tarzan:
Any news on this @bitwalker ?

tarzan:
@bitwalker it would be really great if you can find the time to look at this again. The changes are small, but the impact for our app is big as without it we need to manage the locale of Timex independently instead of it following our global Gettext locale.

tarzan:
@bitwalker any news on this?

Issue #503, opened 2019-01-25
Elixir 1.8 Deprecation warnings


fredr:
### Steps to reproduce

- Install Elixir 1.8.0
- Run test suite

### Description of issue

Elixir 1.8 deprecates pluralized time unit atoms, example error:
```
warning: deprecated time unit: :seconds. A time unit should be :second, :millisecond, :microsecond, :nanosecond, or a positive integer
  (elixir) lib/calendar/iso.ex:703: Calendar.ISO.from_unix/2
  (elixir) lib/calendar/datetime.ex:118: DateTime.from_unix/3
  (elixir) lib/calendar/datetime.ex:166: DateTime.from_unix!/3
  test/timezone_test.exs:102: TimezoneTests."test converting across zone boundaries"/1
  (ex_unit) lib/ex_unit/runner.ex:355: ExUnit.Runner.exec_test/1
  (stdlib) timer.erl:166: :timer.tc/1
  (ex_unit) lib/ex_unit/runner.ex:306: anonymous fn/4 in ExUnit.Runner.spawn_test_monitor/4
```

fireproofsocks:
Did hex pick up a new version of this? The warnings still show up with Timex v 3.5.0.  Thanks for keeping this up to date!

MCMrARM:
Any update on a Hex release? It has been quite a bit of time since this has been merged.

richeterre:
@bitwalker Would you be up for making a release with this?

bitwalker:
Done

PR #504, opened 2019-02-01
Remove deprecations on Elixir 1.8 - Closes Issue #503

scudelletti:
### Summary of changes

Elixir 1.8 deprecates [System.cwd](https://hexdocs.pm/elixir/1.8.1/System.html#cwd/0) and changes the type [time_unit](https://hexdocs.pm/elixir/System.html#t:time_unit/0).

This PR aims to fix those warnings while keeping the old function behavior functional. It should close Issue #503 .

Feel free to reject the PR for any reason(no hard feelings) just let me know why so I can learn from it.
I'm gonna be away from from computer on the next 2 weeks so I won't be able to do any updates during this time.

Thank you for maintaining this library it's awesome! 👍 💐 
### Checklist

### DIFF
--- a/lib/timex/types.ex
+++ b/lib/timex/types.ex
@@ -29,6 +29,7 @@ defmodule Timex.Types do
   @type weekday_name :: :monday | :tuesday | :wednesday | :thursday | :friday | :saturday | :sunday
   @type shift_units :: :milliseconds | :seconds | :minutes | :hours | :days | :weeks | :years
   @type time_units :: :microseconds | :milliseconds | :seconds | :minutes | :hours | :days | :weeks | :years
+  @type second_time_units :: :second | :millisecond | :microsecond | :nanosecond | :seconds | :milliseconds | :microseconds | :nanoseconds
   @type time :: { hour, minute, second }
   @type microsecond_time :: { hour, minute, second, microsecond | microseconds}
   @type date :: { year, month, day }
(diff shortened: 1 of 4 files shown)


lib/timex/types.ex
```diff
@@ -29,6 +29,7 @@ defmodule Timex.Types do
   @type weekday_name :: :monday | :tuesday | :wednesday | :thursday | :friday | :saturday | :sunday
   @type shift_units :: :milliseconds | :seconds | :minutes | :hours | :days | :weeks | :years
   @type time_units :: :microseconds | :milliseconds | :seconds | :minutes | :hours | :days | :weeks | :years
+  @type second_time_units :: :second | :millisecond | :microsecond | :nanosecond | pos_integer() | :seconds | :milliseconds | :microseconds | :nanoseconds
```
ckhrysze: I'm probably missing something, but where does the pos_integer function come from?

scudelletti: Good catch! I got it from [time_unit](https://hexdocs.pm/elixir/System.html#t:time_unit/0) that is used on `System.convert_time_unit/3`.

But seems that [DateTime.from_unix!](https://github.com/elixir-lang/elixir/blob/v1.8.1/lib/elixir/lib/calendar/datetime.ex#L165) does not work with positive integers despite its typespec.

Fixed. Thank you!

williamweckl:
Thanks for the great work!
Any expectations to release that?

PR #509, opened 2019-02-12
Support time units in singular

vorce:
### Summary of changes

Elixir 1.8 emits warnings for using plural time units (ie for `:seconds`), see: #503 
When working with Elixir 1.8 and Timex it's a bit confusing when to use what form of time units.

This PR adds support for singular time units for `Timex.diff` and `Timex.compare`. I am not sure about `shift` - should I change it to also accept singular?

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -166,9 +166,8 @@ defmodule Timex do
 
   Converts the given Unix time to DateTime.
 
-  The integer can be given in different units(:seconds, :milliseconds, :microseconds, :nanoseconds
-  and also units described in `System.convert_time_unit/3`) and it will be converted to
-  microseconds internally. Defaults to `:second`.
+  The integer can be given in different units according to `System.convert_time_unit/3`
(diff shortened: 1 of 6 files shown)


vorce:
I guess the build fails because there are failing tests in master this is branched off of.

lib/timex.ex
```diff
@@ -1026,11 +1045,11 @@ defmodule Timex do
 
     case granularity do
       :duration -> Duration.from_seconds(div(a - b, 1_000 * 1_000))
-      :microseconds -> a - b
-      :milliseconds -> div(a - b, 1_000)
-      :seconds -> div(a - b, 1_000 * 1_000)
-      :minutes -> div(a - b, 1_000 * 1_000 * 60)
-      :hours -> div(a - b, 1_000 * 1_000 * 60 * 60)
+      µs when µs in [:microseconds, :microsecond] -> a - b
```
bitwalker: Can you change this to `us` rather than using the unicode symbol? Thanks

bitwalker:
Thanks!

PR #536, opened 2019-06-01
Support for tzdata 1.0.0

naps62:
This seems to be completely backwards-compatible, and allowing for both versions may be useful to allow people to upgrade other packages


morgz:
Any timeline to get this integrated?

PR #545, opened 2019-07-09
Clearer distinction & examples of .format and .format!

topherhunt:
### Summary of changes

Howdy! I noticed that the [Hexdocs examples for format/2 and format/3](https://hexdocs.pm/timex/Timex.html#format/2) in fact use `.format!` instead of `.format`, and figured that it might be easier to understand if the function called in the example is the same as the function being documented. I also updated the `.format!/2` and `/3` docs to clarify that the return value is tuple-less in case it's not obvious.

### Checklist

- [x] New functions have typespecs, changed functions were updated (n/a)
- [x] Tests were added or updated to cover changes (n/a)
- [x] Commits were squashed into a single coherent commit (n/a)
- [x] Notes added to CHANGELOG file which describe changes at a high-level (n/a)

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -211,12 +211,12 @@ defmodule Timex do
   ## Examples
 
       iex> date = ~D[2016-02-29]
-      ...> Timex.format!(date, "{YYYY}-{0M}-{D}")
-      "2016-02-29"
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks!

PR #559, opened 2019-10-05
Fix not allowed value for seconds

### LINKED ISSUE #506: 60 not allowed as a value for seconds
### Steps to reproduce

```elixir
# parsing 59 as a second value works for the default formatter
Timex.parse("59", "{s}")
# {:ok, ~N[0000-01-01 00:00:59]}

# parsing 60 as a second value does not work for the default formatter
Timex.parse("60", "{s}")
# {:error, "Expected `second` at line 1, column 1."}

# parsing 59 as a second value works for the strftime formatter
Timex.parse("59", "%s", :strftime)
# {:ok, #DateTime<1970-01-01 00:00:59Z>}

# parsing 60 as a second value doesn't error with the strftime formatter but it does
# assume it is the beginning of the next minute
Timex.parse("60", "%s", :strftime)
# {:ok, #DateTime<1970-01-01 00:01:00Z>}

# This is compounded by Elixir itself not parsing leap seconds
~N[2016-12-31 23:59:59]
# ~N[2016-12-31 23:59:59]
~N[2016-12-31 23:59:60]
# ** (ArgumentError) cannot parse "2016-12-31 23:59:60" as naive datetime, reason: :invalid_time
```

### Description of issue

I'm trying to allow for leap seconds in time strings but I can't see a way to. The docs for both the [default formatter](https://hexdocs.pm/timex/Timex.Format.DateTime.Formatters.Default.html#module-time) and the [strftime formatter](https://hexdocs.pm/timex/Timex.Format.DateTime.Formatters.Strftime.html#module-time) say the second value can go up to 60.

Thoughts on what's going on? Am I doing something wrong?

lucazulian:
[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

fix #506

### DIFF
--- a/lib/parse/datetime/parser.ex
+++ b/lib/parse/datetime/parser.ex
@@ -402,7 +402,15 @@ defmodule Timex.Parse.DateTime.Parser do
         %{date | :minute => value}
 
       :sec ->
-        %{date | :second => value}
+        case value do
+          60 ->
+            date
+            |> Timex.to_naive_datetime()
+            |> (&%{&1 | second: 0}).()
+            |> Timex.shift([minutes: 1])
+          value ->
+            %{date | :second => value}
+        end
 
       :sec_fractional ->
         case value do
--- a/lib/parse/datetime/parser.ex
+++ b/lib/parse/datetime/parser.ex
@@ -406,7 +406,7 @@ defmodule Timex.Parse.DateTime.Parser do
           60 ->
             date
             |> Timex.to_naive_datetime()
-            |> (&%{&1 | second: 0}).()
+            |> Timex.set(second: 0)
             |> Timex.shift([minutes: 1])
           value ->
             %{date | :second => value}
--- a/lib/parse/datetime/parsers.ex
+++ b/lib/parse/datetime/parsers.ex
@@ -181,7 +181,7 @@ defmodule Timex.Parse.DateTime.Parsers do
 
   def second(opts \\ []) do
     Helpers.integer(opts)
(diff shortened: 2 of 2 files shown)


lucazulian:
News about that PR?

lib/parse/datetime/parser.ex
```diff
@@ -318,7 +318,16 @@ defmodule Timex.Parse.DateTime.Parser do
       hour when hour in [:hour24, :hour12] ->
         %{date | :hour => value}
       :min       -> %{date | :minute => value}
-      :sec       -> %{date | :second => value}
+      :sec       ->
+        case value do
+          60 ->
+            date
+            |> Timex.to_naive_datetime()
+            |> (&%{&1 | second: 0}).()
```
bitwalker: I would prefer this line to be either `Map.put(:second, 0)` or `Timex.set(second: 0)`, as it is clearer what is happening. Either that or break up the pipeline. For readability, I prefer to never use anonymous functions in pipelines, especially the shorthand syntax, since it is unnecessarily noisy compared to `fn ... -> ... end`.

Other than that, I think this looks good, and can be merged once that change is made.

lucazulian: Done!

lucazulian: @bitwalker can you merge this please?

bitwalker:
You’ll have to rebase on master

lucazulian:
> You’ll have to rebase on master

Done, thank you!

Issue #494, opened 2018-12-21
doesnt parse week of year the say way it formats


atomkirk:
```
> Timex.format!(~D[2018-01-01], "%U", :strftime) |> Timex.parse("%U", :strftime)
{:error, "Expected `week of year` at line 1, column 1."}
```

seems like parse should be an exact reverse of format

bitwalker:
Fixed in master

PR #562, opened 2019-11-01
fixed: format 'week of year' must return a value between 1..53.

diasbruno:
### Summary of changes

As reported on #494, formatting a date and parsing the results with the same format string, must produce a valid date (in this case the week of year).

```ex
Timex.format!(~D[2018-01-01], "%U", :strftime) |> Timex.parse("%U", :strftime)
{:error, "Expected `week of year` at line 1, column 1."}
```

NOTE: I still need to check if there are more thing to make this right. Appreciate some help.

### Checklist


bitwalker:
This ended up getting fixed during recent refactoring and bug chasing, so I'm going to close this. Thanks for the contribution, it's much appreciated!

PR #563, opened 2019-11-05
Implement beginning_of_iso_week

tomasz-tomczyk:
### Summary of changes

I wanted to find out what is the first day of a given week of the year. I couldn't find this functionality in Timex at the moment (although I may be missing something obvious!) and when I tried to implement it on top of it, I found a couple of edge cases and thought it might something of value to others.

Let me know if this is of use and any feedback to make it better! Thanks!

Thanks to @Vereis for helping out.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


tomasz-tomczyk:
Realised I should implement the protocol for remaining types - will update soon

PR #568, opened 2019-12-30
fix `Timex.week_of_month/3` to support leap weeks

mkaszubowski:
### Summary of changes

We have to handle a special case when iso_week for a date
returns the first week of a following year.
For example: `iso_week(2019, 12, 30)` is `{2020, 1}`, which results in
this function returning -46 for this date. Other examples are:
2024-12-31, 2029-12-31, 2030-12-31.

More details here: https://en.wikipedia.org/wiki/ISO_week_date

### Checklist


bitwalker:
This is superceded by #616

Issue #579, opened 2020-02-12
Is a repository cleanup welcome?


qhwa:
First of all, thank you for the excellent work on this project. I used it in many projects of mine and really enjoyed it.

It seems that codes are not formatted and the specifications are broken.

### Steps to reproduce

- mix format --check-formatted lib/**/*.ex
- mix dialyzer

### Description of issue

I saw there are some issues open and thought maybe I can help. After cloning the project and running it, I think that maybe a cleanup is a good start, future updates will benefit from it.

But I am not sure whether this is a good idea or not, maybe it will cause a lot of conflicts with other developing branches (although there seem to be no active developing branches). So I open this issue to ask if code formatting with `mix format` and fixing of specifications welcome? If so, I will start work on the master branch and open a pull request later.

What do you think? @bitwalker 

UPDATE: Sorry I was reading the codes on an outdated revision. I see most of the files in the master branch has already been formatted. That's awesome. Things get a lot easier based on your work.

bitwalker:
Additional help keeping the codebase tidy and building clean is always welcome :)

PR #581, opened 2020-02-13
format code

qhwa:
### Summary of changes

The project is formatted with `mix format`,  for a better future developing experience.  

[related issue](https://github.com/bitwalker/timex/issues/579)

- [x] All the codes (`*.ex`, `*.exs`) are formatted with `mix format`;
- [x] Travis CI scripts are updated, with `mix format --check-formatted` added in scripts when running on Elixir 1.10.x, in order to check future integrations;
- [x] Travis CI `elixir` and `otp_release` configurations are updated to the latest stable version (22.2) in order to stick with the latest Elixir version and formatting rules. 

Specification fixes will follow in another pull request.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,23 +5,34 @@ os:
   - macos
   - linux
 elixir:
-  - 1.6
-  - 1.7
-  - 1.8
-  - 1.9
+  - '1.6'
+  - '1.7'
+  - '1.8'
+  - '1.9'
+  - '1.10'
 otp_release:
-  - 19.3
-  - 20.3
-  - 21.0
+  - '19.3'
+  - '20.3'
+  - '21.0'
+  - '22.2'
 matrix:
   exclude:
     - os: windows
-    - elixir: 1.8
-      otp_release: 19.3
-    - elixir: 1.9
-      otp_release: 19.3
+    - elixir: '1.6'
+      otp_release: '22.2'
+    - elixir: '1.7'
+      otp_release: '22.2'
+    - elixir: '1.8'
+      otp_release: '19.3'
+    - elixir: '1.9'
+      otp_release: '19.3'
+    - elixir: '1.10'
+      otp_release: '19.3'
+    - elixir: '1.10'
+      otp_release: '20.3'
 script:
   - if [[ "$TRAVIS_OS_NAME" = 'windows' ]]; then ./bin/build_win32; else ./bin/build; fi
+  - if [[ "$TRAVIS_ELIXIR_VERSION" = 1.10* ]]; then mix format --check-formatted; fi
 notifications:
   recipients:
     - paulschoenfelder@fastmail.com
--- a/lib/parse/datetime/parser.ex
+++ b/lib/parse/datetime/parser.ex
@@ -407,7 +407,8 @@ defmodule Timex.Parse.DateTime.Parser do
             date
             |> Timex.to_naive_datetime()
             |> Timex.set(second: 0)
-            |> Timex.shift([minutes: 1])
+            |> Timex.shift(minutes: 1)
+
           value ->
             %{date | :second => value}
         end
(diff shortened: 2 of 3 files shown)


qhwa:
CI is broken, fixing it.

lib/parse/datetime/parser.ex
```diff
@@ -407,7 +407,8 @@ defmodule Timex.Parse.DateTime.Parser do
             date
             |> Timex.to_naive_datetime()
             |> Timex.set(second: 0)
-            |> Timex.shift([minutes: 1])
+            |> Timex.shift(minutes: 1)
+
```
qhwa: This line is changed by `mix format`.

.travis.yml
```diff
@@ -5,23 +5,34 @@ os:
...
+      otp_release: '22.2'
+    - elixir: '1.7'
+      otp_release: '22.2'
+    - elixir: '1.8'
+      otp_release: '19.3'
+    - elixir: '1.9'
+      otp_release: '19.3'
+    - elixir: '1.10'
+      otp_release: '19.3'
+    - elixir: '1.10'
```
qhwa: `1.10` is different from `'1.10'` in yaml.

.travis.yml
```diff
@@ -5,23 +5,34 @@ os:
...
   exclude:
     - os: windows
-    - elixir: 1.8
-      otp_release: 19.3
-    - elixir: 1.9
-      otp_release: 19.3
+    - elixir: '1.6'
+      otp_release: '22.2'
+    - elixir: '1.7'
+      otp_release: '22.2'
```
qhwa: These elixir / OTP groups failed in CI, so are disabled:

* 1.6 / 22
* 1.7 / 22

.travis.yml
qhwa: OTP 21+ is required by Elixir 1.10.

.travis.yml
```diff
@@ -5,23 +5,34 @@ os:
...
 elixir:
-  - 1.6
-  - 1.7
-  - 1.8
-  - 1.9
+  - '1.6'
+  - '1.7'
+  - '1.8'
+  - '1.9'
+  - '1.10'
```
qhwa: `1.10` is different from `'1.10'` (notice the single quote mark `'`).

.travis.yml
```diff
@@ -5,23 +5,34 @@ os:
...
+      otp_release: '19.3'
+    - elixir: '1.9'
+      otp_release: '19.3'
+    - elixir: '1.10'
+      otp_release: '19.3'
+    - elixir: '1.10'
+      otp_release: '20.3'
 script:
   - if [[ "$TRAVIS_OS_NAME" = 'windows' ]]; then ./bin/build_win32; else ./bin/build; fi
+  - if [[ "$TRAVIS_ELIXIR_VERSION" = 1.10* ]]; then mix format --check-formatted; fi
```
qhwa: Because `mix format` uses different rules on different Elixir versions, here we just use the latest Elixir version, which is `1.10.1` currently.

It may not be the best choice to hardcode a version number here, but I didn't find any other way better.

.travis.yml
bitwalker: I'm not sure I understand what the difference is supposed to be when quoted vs not quoted. Can you provide a bit more detail on the purpose of that change?

qhwa: `1.10` without quote will be parsed as a float number and turned into `1.1`. This was the reason why [this travis build](https://travis-ci.org/bitwalker/timex/builds/649746838?utm_medium=notification&utm_source=github_status) was using Elixir `1.1`.  :rofl:

qhwa:
@bitwalker Is there anything else I need to update?

bitwalker:
Thanks!

PR #583, opened 2020-02-23
Run CI against 1.10

RicardoTrindade:
### Summary of changes

Added elixir 1.10 and otp 22 to travis.yml so CI runs against latest elixir version as well
### Checklist


bitwalker:
We've migrated to Cirrus, which now builds against latest

PR #589, opened 2020-03-13
Modify default precision

### LINKED ISSUE #588: Timex.to_datetime truncates microseconds
### Steps to reproduce

```
> Timex.to_datetime({{2017, 1, 1}, {0, 0, 0, 0}})
~U[2017-01-01 00:00:00Z]
> Timex.to_datetime({{2017, 1, 1}, {0, 0, 0, 1}})
~U[2017-01-01 00:00:00.000001Z]
```

### Description of issue

I would expect the result of the first call to be `~U[2017-01-01 00:00:00.000000Z]`.

lucazulian:
fix #588 
I'm really not sure that this is the right way to solve the issue and I am not sure that it is an issue.
The default precision in based on the value required for that precision; with this PR I modify this concept.

### DIFF
--- a/lib/datetime/helpers.ex
+++ b/lib/datetime/helpers.ex
@@ -119,7 +119,7 @@ defmodule Timex.DateTime.Helpers do
   end
 
   # Input precision of -1 means it should be recalculated based on the value
-  def construct_microseconds(0, -1), do: {0, 0}
+  def construct_microseconds(0, -1), do: {0, 6}
   def construct_microseconds(0, p), do: {0, p}
   def construct_microseconds(n, -1), do: {n, precision(n)}
   def construct_microseconds(n, p), do: {to_precision(n, p), p}
--- a/lib/time/duration.ex
+++ b/lib/time/duration.ex
@@ -54,7 +54,7 @@ defmodule Timex.Duration do
 
       iex> d = %Timex.Duration{megaseconds: 0, seconds: 4000, microseconds: 0}
       ...> Timex.Duration.to_time(d)
-      {:ok, ~T[01:06:40]}
+      {:ok, ~T[01:06:40.000000]}
 
       iex> d = %Timex.Duration{megaseconds: 1, seconds: 0, microseconds: 0}
       ...> Timex.Duration.to_time(d)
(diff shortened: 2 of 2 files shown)


bitwalker:
Can you explain the motivation behind this change? We explicitly want to preserve precision, or match the precision of inputs, in those cases where a precision was not specified or detectable.

lucazulian:
The motivation behind this is due to the issue #588, and I'm not sure it's the correct way to solve the issue. This is just an attempt, do you have any suggestion? Or can we close the issue as we don't see the need for an extra precision?

mindreader:
A side note about this pr.  I think this pr would also fix a problem I had.

```
> Timex.to_datetime({{2017, 1, 1}, {0, 0, 0,0}}) |> Timex.format("{RFC3339z}")
{:ok, "2017-01-01T00:00:00Z"}
> Timex.to_datetime({{2017, 1, 1}, {0, 0, 0,1}}) |> Timex.format("{RFC3339z}")                                                              
{:ok, "2017-01-01T00:00:00.000001Z"}
```
I don't care if the inspect format truncates, but I've also noticed that several of the formatting functions also truncate.  We had an issue where once in awhile elastic search queries would fail sometimes.  It requires that you specify the format of time stamps in its queries and it either accepts a format with milliseconds or without it.

Once every million queries it would fail, and that would occur because the date we were sending to elasticsearch would not have milliseconds because of this truncation.  Luckily the fix was easy in elasticsearch (tell it to try one format, then the other), but I imagine that kind of fix will not always be easy.

bitwalker:
@mindreader In your example there, the first date/time had no microseconds (and thus zero precision), while the second one has exactly 1 microsecond (and so requires full precision). 

When converting from an Erlang timestamp to an Elxir date/time, we infer the precision based on the precision required to accurately express the input; but Elixir date/times can have values which are more precise than the precision set in the structure itself (i.e. Elixir separates the desired precision from the value, which is then used in printing/etc.). 

So in the case of converting an Erlang timestamp (w/ microseconds) to an Elixir date/time and printing it, the results you see are exactly what we'd expect. In the case where an Elixir date/time is the input though, we use the precision defined in the date/time structure itself, rather than infer from the value. You can force zero precision (or full precision), by setting the precision of the DateTime/NaiveDateTime/etc. structure manually (or by providing the precision during construction, if using the Elixir APIs).

@lucazulian The above is why #589 is not a bug, but expected behavior - a microseconds value of 0 is the default value of a zero-precision value (and requires no microsecond precision to express). Your application code should instead force the precision to 6 after constructing the date/time value if that is what is desired in your application.

Since I don't believe this PR is actually addressing a bug, I'm going to close this for now, but I'm open to continuing the discussion if there is still some question around the behavior here.

PR #595, opened 2020-03-31
Bosnian language translations

almirsarajcic:
This PR includes both Latin and Cyrillic scripts for Bosnian language.


bitwalker:
Thanks!

PR #611, opened 2020-08-29
Add for: Timex.Interval to Enumerable implementation

dstockdale:
I can't find any documentation about `defimpl Enumerable` now requiring the `for:` option to be explicitly declared but adding `for: Timex.Interval` has instantly cured the problem we were having on Elixir 1.9.4 where it was saying that Timex.Enumerble was NOT Enumerable.

I've looked in a bunch of other libs like postgrex and Elixir itself and most modules defining `defimpl` are explicit in this way so seems to be correct but also harmless.

### Summary of changes

### Checklist


bitwalker:
Thanks!

PR #618, opened 2020-10-02
Fix #616 - fix negative week_of_month/3

### LINKED ISSUE #616: Negative week numbers
### Steps to reproduce

- `Timex.week_of_month/1` 
- Issue with dates `~D[2021-01-10]`, `~D[2021-01-17]`, `~D[2021-01-24]`, and `~D[2021-01-31]`.

### Description of issue

- What are the expected results?
Dates aforementioned are supposed to return week numbers 2, 3, 4, 5 respectively. However, it returns -51, -50, -49, -48. For week one of January 2021, it returns the right week number though.

image

- If you are certain this is a bug in Timex, do you have references I
  can use to check the algorithm/logic against? For instance, a well-established
  date/time library in another language, a paper, test suite, etc. It is by no
  means a requirement, but is immensely helpful in fixing bugs, and provides me
  extra data to use when building out the test suite
I'm not sure, sorry!

syukronrm:
### Summary of changes

Fixes issue #616 

Currently `week_of_month/3` relying on `iso_week/3`. But at certain cases `iso_week/3` return different year than the year input, it may return the previous or the next year.

Example case for returning week on the previous year.
```
iex(1)> Timex.iso_week(2021, 1, 1)
{2020, 53}
```
Thus, it leads to:
```
iex(3)> Timex.week_of_month(2021, 1, 10)
-51 # should be 1
```

This is the case for returning the next year.
```
iex(6)> Timex.iso_week(2019, 12, 31)     
{2020, 1}
```
Leads to:
```
iex(7)> Timex.week_of_month(2019, 12, 31)
-46 # should be 6
```
### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -1312,8 +1312,28 @@ defmodule Timex do
   """
   @spec week_of_month(Types.year(), Types.month(), Types.day()) :: Types.week_of_month()
   def week_of_month(year, month, day) when is_date(year, month, day) do
-    {_, week_index_of_given_date} = iso_week(year, month, day)
-    {_, week_index_of_first_day_of_given_month} = iso_week(year, month, 1)
+    next_year = year + 1
+    prev_year = year - 1
+
+    week_index_of_given_date =
+      case iso_week(year, month, day) do
+        {^prev_year, _} ->
+          0
+
+        {^next_year, _} ->
+          {_, week_index_of_given_date_shifted} = iso_week(year, month, day - 7)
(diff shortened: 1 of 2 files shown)


syukronrm:
I believe this also covers PR #568.

bitwalker:
The behaviour you specified here is expected - that is how ISO weeks are defined, they don't necessarily perfectly overlap with calendar weeks/year.

bitwalker:
Sorry, I'm dumb - missed that this wasn't changing the ISO week behavior.

PR #619, opened 2020-10-02
set MIX_HOME in CI

davemenninger:
### Summary of changes

Mix fails to get dependencies in Cirrus CI, using `elixir:1.10`:

```
(Mix) Could not find an SCM for dependency :tzdata from Timex.Mixfile
```

Adding `MIX_HOME` to the ENV makes it work:

* https://gitlab.com/gitlab-org/security-products/analyzers/sobelow/-/merge_requests/2
* https://pre.hn/elixir-on-heroku-with-docker
* https://cirrus-ci.com/task/6461184138280960

### Checklist

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -36,7 +36,9 @@ task:
     brew install erlang elixir
   mix_cache:
     folder: deps
-    fingerprint_script: cat mix.lock
+    fingerprint_script:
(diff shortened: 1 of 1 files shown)


davemenninger:
Huh.  I don't know why it works for me: https://cirrus-ci.com/task/6461184138280960 but doesn't work here.

PR #620, opened 2020-10-02
fixes for caching in Cirrus CI

davemenninger:
### Summary of changes

`mix compile` needs Hex (and Rebar) to build dependencies:

```
Could not find Hex, which is needed to build dependency :tzdata
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :tzdata from Timex.Mixfile
```

We can add it to a cache ( as in the Linux task in this PR ), or we can install it every time ( as in the MacOS task in this PR ).

* https://hexdocs.pm/mix/master/Mix.Tasks.Local.Hex.html#content
* https://hexdocs.pm/mix/master/Mix.Tasks.Local.Rebar.html#content
* https://cirrus-ci.org/guide/writing-tasks/#cache-instruction

### Checklist

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,16 +11,23 @@ task:
     image: elixir:1.10
   env:
       MIX_ENV: 'test'
+      MIX_HOME: '/opt/mix'
       LC_ALL: 'en_US.UTF-8'
       LANG: 'en_US.utf8'
   mix_cache:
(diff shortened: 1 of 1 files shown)


bitwalker:
Closing since we're not using Cirrus anymore

PR #633, opened 2020-12-28
fix: czech relative times

segeda:
### Summary of changes

In czech `next week` is translate to `příští týden`. `Next` is not the same as `futured`

Viz Google Translator https://translate.google.com/?sl=en&tl=cs&text=future%20month%0Anext%20month&op=translate


bitwalker:
Thanks!

PR #635, opened 2021-01-12
Fix elixir github workflow

lucazulian:
### Summary of changes

I notice a error on `elixir` github workflow so I fixed with matrix strategy
```
The workflow is not valid. .github/workflows/elixir.yml (Line: 12, Col: 27): A sequence was not expected .github/workflows/elixir.yml (Line: 13, Col: 24): A sequence was not expected
```

### Checklist

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -3,14 +3,23 @@ on:
   pull_request:
 
 jobs:
-  test:
+  build:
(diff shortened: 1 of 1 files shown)


lucazulian:
I would like to add dialyzer check, maybe in another PR, what do you think?

bitwalker:
I ended up fixing this during the course of some other changes, thanks for the contribution though!

PR #636, opened 2021-01-28
Support one-digit hour when parsing timezone offset

k4ju:
Recently Google Calendar API started to respond with timezones in the form of `GMT-/+H:MM` — not preceding the one-digit hour mark with `0`. This is not recognised by timex resulting in **FunctionClauseError** in Timex.Timezone.parse_offset/1. While the `GMT-0H:MM` format is perfectly supported there should be no risk in adding this variation.


k4ju:
🙇

PR #639, opened 2021-02-23
Fix Markdown formatting in moduledoc

oleks:
### Summary of changes

Prior to this, parts of these sentences would be incorrectly typeset
as code.

### Checklist


bitwalker:
Thanks!

PR #640, opened 2021-02-23
Fix zero behavior for default duration formatter

aj-foster:
### Summary of changes

This PR changes the behavior of the default duration formatter for a zero-length duration. Previously it would report `P`, and now it reports `PT0S`. The use of seconds, as opposed to another unit, was arbitrary.

**Why**: [Second-hand reading](https://stackoverflow.com/a/29824127/4250566) of the ISO 8601 spec* says...

>If the number of years, months, days, hours, minutes or seconds in any of these expressions equals zero, the number and the corresponding designator may be absent; however, at least one number and its designator shall be present.

As a result, `P` is considered invalid whereas `PT0S` is an equivalent, valid expression.

*I haven't paid to read the paper myself.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Potentially Breaking
 - `Timex.set/2` now sets microseconds when setting `:time` from a `%Time{}` struct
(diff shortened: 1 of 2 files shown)


bitwalker:
Thanks!

PR #655, opened 2021-03-26
fix: converting with empty time zone

parallel588:
fixed `Timex.Timezone.name_of` for empty value

### DIFF
--- a/lib/timezone/timezone.ex
+++ b/lib/timezone/timezone.ex
@@ -224,6 +224,8 @@ defmodule Timex.Timezone do
       when sign in [?+, ?-],
       do: "Etc/UTC" <> <<sign::utf8, hh::binary, ?:, mm::binary>>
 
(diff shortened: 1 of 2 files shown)


lib/timezone/timezone.ex
```diff
@@ -224,6 +224,8 @@ defmodule Timex.Timezone do
       when sign in [?+, ?-],
       do: "Etc/UTC" <> <<sign::utf8, hh::binary, ?:, mm::binary>>
 
+  def name_of(<<tz::binary-size(0)>>), do: {:error, :time_zone_not_found}
```
bitwalker, line 227: You can change this to be simply `def name_of(""), do: {:error, :time_zone_not_found}`

bitwalker (changes requested):
Thanks! One small change before we merge this though.

bitwalker (approved):

PR #663, opened 2021-04-22
Prevents :time_zone_not_found when no space  at the end on timezone value.

paridin:
### Summary of changes

After update to 3.7.5 I started to receive `:time_zone_not_found`, I detect the problem was related to the value obtained from the reading of the `/etc/timezone` file.

```elixir
iex(web@da370b35f69c)1> Timex.local()
{:error, :time_zone_not_found}
```

```bash
# I have the value `America/MexicoCity` **no spaces at the end** on /etc/timezone file
echo "America/Mexico_City" > /etc/timezone
```

After read the code I saw there is a space during the call of `String.trim(name, " \n")`; 
My next step was update the value from the timezone file I just add the missing space and it worked. 

```bash
echo "America/Mexico_City " > /etc/timezone
```

Then I decided update to `String.replace(name, ~r/[\s\n]/, "")` to keep both cases with space(s) and with no space(s).

I'm unsure if checklist is required to this particular change because it only applies to private function. Let me know if they are required.

### Checklist

### DIFF
--- a/lib/timezone/local.ex
+++ b/lib/timezone/local.ex
@@ -229,7 +229,7 @@ defmodule Timex.Timezone.Local do
   defp read_timezone_data(_, @_ETC_TIMEZONE) do
     case File.read(@_ETC_TIMEZONE) do
       {:ok, name} ->
-        {:ok, String.trim(name, " \n")}
+        {:ok, String.replace(name, ~r/[\s\n]/, "")}
 
       {:error, _} ->
         nil
--- a/lib/timezone/local.ex
+++ b/lib/timezone/local.ex
@@ -229,7 +229,7 @@ defmodule Timex.Timezone.Local do
   defp read_timezone_data(_, @_ETC_TIMEZONE) do
     case File.read(@_ETC_TIMEZONE) do
       {:ok, name} ->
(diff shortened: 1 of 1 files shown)


bitwalker:
So the Olson timezone names don't contain spaces at all, instead they use `_` where a space would be, so that's the name you would want to use in your `/etc/timezone` file. 

As for the proposed change here: I would prefer to not modify the value that someone puts in `/etc/timezone` because it will make troubleshooting very painful. We currently trim leading/trailing spaces and newlines just because those frequently are accidentally present when writing a file (particularly trailing newlines), but we don't want to modify anything else about the value. Instead, let's change `String.trim(value, " \n")` to just `String.trim(value)`, which does what we want (remove leading/trailing whitespace).

bitwalker (changes requested):
Let's change this to just `{:ok, String.trim(name)}`

paridin:
Thanks for the review, I applied the suggestion.

edwardzhou:
@bitwalker It is totally a blocker issue. It would be appreciated if you can quick review, merge and release a new version.
currently, the work around is config timezone in config.exs
config :timex, local_timezone: "Asia/Shanghai"

bitwalker (approved):

PR #671, opened 2021-06-02
Set correct day when year or month is changed too

oliver-kriska:
### Summary of changes
 
- when you call `Timex.set/2` function from date which has less days than new month it should apply month or year changes before day changes so validation will be correct. So options have to be in exact order: year, month, day
 - added alias for Helpers module

Example before this fix:

```Elixir
Timex.now() |> Timex.set(day: 31, month: 5)
~U[2021-05-30 17:49:07.917681Z]
Timex.now() |> Timex.set(month: 5, day: 31)
~U[2021-05-31 17:49:10.264841Z]
```
There you can see correct date is only when I change order of options. 

### Checklist

### DIFF
--- a/lib/datetime/helpers.ex
+++ b/lib/datetime/helpers.ex
@@ -145,4 +145,28 @@ defmodule Timex.DateTime.Helpers do
       new_p
     end
   end
+
+  def sort_options(options) when is_list(options) do
+    options =
+      case Keyword.pop(options, :day) do
+        {nil, options} -> options
+        {day, opts} -> Keyword.put(opts, :day, day)
+      end
+
+    options =
+      case Keyword.pop(options, :month) do
+        {nil, options} -> options
+        {month, opts} -> Keyword.put(opts, :month, month)
+      end
+
+    options =
+      case Keyword.pop(options, :year) do
+        {nil, options} -> options
+        {year, opts} -> Keyword.put(opts, :year, year)
+      end
+
+    options
+  end
+
+  def sort_options(options), do: options
 end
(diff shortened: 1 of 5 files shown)


oliver-kriska:
I did some cosmetic changes so there 3 commits. In case it will be ok, I can make new PR with one commit.

erik-mocny (approved):

lib/datetime/helpers.ex
```diff
@@ -145,4 +145,28 @@ defmodule Timex.DateTime.Helpers do
       new_p
     end
   end
+
+  def sort_options(options) when is_list(options) do
```
bitwalker, line 149: I would instead do something like this:

```elixir
@set_option_priority %{
  year: 1,
  month: 2,
  day: 3,
  hour: 4,
  minute: 5,
  second: 6,
}

def sort_options(options) when is_list(options) do
  Enum.sort_by(options, fn {k, _} -> Map.get(@set_option_priority, k, 99) end)
end
```

This is a lot easier to maintain and a lot more readable, and ensures that other options are always sorted _after_ the explicitly prioritized attributes. Really we should define the priority for all valid options, but these will do for now.

bitwalker (changes requested):
While this behavior _is_ documented, I think this is a worthwhile change to implement. Two things:

- I left one suggested change I'd like to have you make
- Could you also rebase/squash your commits? You don't have to break up the cosmetic changes at this point (please do in the future though)

Thanks for the PR!

oliver-kriska:
I added your suggestion code, all tests passed. It looks now better. Thanks for good suggestion. 

Sorry but I don't understand what you want with this:
> Could you also rebase/squash your commits? You don't have to break up the cosmetic changes at this point (please do in the future though)

I updated my main branch and now it's code base same as in your master, plus my changes. I don't like doing hard reset and creating one commit. Because it can be done when this PR will be merged. Because we would lost history about the changes and I'm not sure about this PR, maybe it will just update PR or I would have to create new one after reset and squash commits. Maybe I'm wrong so sorry and please give me a advice. 
CleanShot 2021-08-03 at 10 19 22

Zarathustra2:
Any chance this can be merged?

PR #672, opened 2021-06-15
Fixing ignored overlapping specs for Time.t()

bobkocisko:
### Summary of changes

Based on my tests, Dialyzer is ignoring all the duplicate specs that have the same arity in Timex.ex and therefore reporting errors every time I try to use the functions with %Time{} arguments.

### Checklist

- [ ] Tests were added or updated to cover changes (**I don't see a clear way to add dialyzer tests**)
- [ ] Notes added to CHANGELOG file which describe changes at a high-level (**Is this necessary for such a small commit?**)


bitwalker:
Looks good, thanks!

PR #687, opened 2021-09-18
add timezone to the set_options type

mwhitworth:
### Summary of changes

### Checklist


fceruti:
I'm having this issue. Dyalizer is going crazy everywhere in my code base because of this. 

Funny thing is, it still works as expected, even thou the specs are incomplete.

fceruti:
If you are here before this is merged, here's an elixir only approach to setting a timezone to a utc_datetime

```elixir
  def set_timezone(utc_dt, timezone) do
    %NaiveDateTime{
      year: utc_dt.year,
      month: utc_dt.month,
      day: utc_dt.day,
      hour: utc_dt.hour,
      minute: utc_dt.minute,
      second: utc_dt.second
    }
    |> DateTime.from_naive!(timezone)
  end
  ```

bitwalker (approved):

Issue #674, opened 2021-07-02
Timex.parse() does not handle ambiguous times in 3.7.5


billperegoy:
### Steps to reproduce

```
    format_string = "{ISOdate}T{ISOtime}-{Zname}"
    ambiguous_dst_dt = "2021-11-07T01:00:00.000-America/New_York"

    Timex.parse(ambiguous_dst_dt, format_string)
```

In Timex 3.7.3, this returns:

```
{:ok,
 #<Ambiguous(#DateTime<2021-11-07 01:00:00-04:00 EDT America/New_York> ~ #DateTime<2021-11-07 01:00:00-05:00 EST America/New_York>)>}
 ```

In Timex 3.7.5, this crashes with this error:

```
     The following arguments were given to Timex.Parse.DateTime.Parser.validate_datetime/1:
     
         # 1
         #<Ambiguous(#DateTime<2021-11-07 01:00:00-04:00 EDT America/New_York> ~ #DateTime<2021-11-07 01:00:00-05:00 EST America/New_York>)>
     
     Attempted function clauses (showing 1 out of 1):
     
         defp validate_datetime(%{year: y, month: m, day: d} = datetime)
     
     code: Timex.parse(ambiguous_dst_dt, format_string)
     stacktrace:
       (timex 3.7.5) lib/parse/datetime/parser.ex:230: Timex.Parse.DateTime.Parser.validate_datetime/1
       (timex 3.7.5) lib/parse/datetime/parser.ex:215: Timex.Parse.DateTime.Parser.apply_directives/3
       (timex 3.7.5) lib/parse/datetime/parser.ex:95: Timex.Parse.DateTime.Parser.parse!/3
       (timex 3.7.5) lib/parse/datetime/parser.ex:58: Timex.Parse.DateTime.Parser.parse/3
```

### Description of issue

I'm not sure if this was intentionally removed or not. In looking at the types on `Timex.Parse.DateTime.Parser.parse/2`, I notice that Ambiguous times were ever in the return type. So maybe this was never intended to be a feature of `Timex.parse`?

bitwalker:
This is an unintentional breakage, we should be handling this properly, or at least raising a proper error if for some reason we can't proceed with the ambiguity.

PR #689, opened 2021-10-02
Fix specs for parse to allow AmbiguousDateTime

billperegoy:
### Summary of changes
When issue:
https://github.com/bitwalker/timex/issues/674

was addressed to allow the parse function to return AmbiguousDateTime,
the specs were not updated and this can cause dialyzer errors in code
that uses this function.

### Checklist

### DIFF
--- a/lib/timex.ex
+++ b/lib/timex.ex
@@ -508,9 +508,10 @@ defmodule Timex do
       true
 
   """
-  @spec parse(String.t(), String.t()) :: {:ok, DateTime.t() | NaiveDateTime.t()} | {:error, term}
+  @spec parse(String.t(), String.t()) ::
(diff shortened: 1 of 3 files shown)


lib/timex.ex
```diff
@@ -1060,7 +1072,11 @@ defmodule Timex do
 
   and the result will be an integer value of those units or a Duration.
   """
-  @spec diff(Time.t() | Comparable.comparable(), Time.t() | Comparable.comparable(), Comparable.granularity()) ::
+  @spec diff(
+          Time.t() | Comparable.comparable(),
+          Time.t() | Comparable.comparable(),
+          Comparable.granularity()
+        ) ::
```
billperegoy, line 1079: These are unrelated changes that `mix format` made.

bitwalker (approved):

PR #690, opened 2021-10-04
Add latvian translations

ramuuns:
saw that those were missing, and was keen on using them for stuff, so I figured I'd be nice and make them a thing.

Some things of note.
1) Latvians don't really use the 1000 separator, it's not really a thing over here
2) You might notice that the abbreviations are somewhat unusual (e.g. march and may aren't abbreviated) that's because apparently according to the official language center documents those months _aren't_ abbreviated
(see http://www.vvc.gov.lv/advantagecms/export/docs/noderigi/saisinajumi/saisinajumi-1.pdf) additionally days of the week are generally abbreviated to a single letter, and second one is added only to the second day of that week that begins with said letter (e.g. S - Sestdiena (Saturday)  and Sv - Svētdiena (Sunday)

### Summary of changes

### Checklist

- [n/a ] New functions have typespecs, changed functions were updated
- [n/a ] Same for documentation, including moduledocs
- [n/a] Tests were added or updated to cover changes
- [✅] Commits were squashed into a single coherent commit
- [X] Notes added to CHANGELOG file which describe changes at a high-level

### DIFF
--- a/priv/translations/lv/LC_MESSAGES/day_periods.po
+++ b/priv/translations/lv/LC_MESSAGES/day_periods.po
@@ -1,0 +1,27 @@
+# # `msgid`s in this file come from POT (.pot) files.
+# #
+# # Do not add, change, or remove `msgid`s manually here as
+# # they're tied to the ones in the corresponding POT file
+# # (with the same domain).
+# #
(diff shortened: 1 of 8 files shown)


bitwalker (approved):
Thanks so much!

PR #698, opened 2021-12-30
Fix lformat wrong on AM/PM cases in locale tr

thaiyojee:
### Summary of changes

Fix the translation on the day_periods domain in the Turkish language.
I try a test case like this and it's failed like
```
Timex.lformat!(~N[2021-12-28 05:15:30], "%d %b %I:%M%P", "tr", :strftime)
<<50, 56, 32, 65, 114, 97, 32, 48, 53, 58, 49, 53, 246, 246>>
```
Because the character %P/%p on the Turkish translation in the source code may be wrong encoding (display like a diamond with `?` mark when I pull from the repo)

### Checklist

### DIFF
--- a/priv/translations/tr/LC_MESSAGES/day_periods.po
+++ b/priv/translations/tr/LC_MESSAGES/day_periods.po
@@ -22,16 +22,16 @@ msgstr ""
 
 #: lib/l10n/translator.ex:266
 msgid "AM"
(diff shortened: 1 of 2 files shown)


bitwalker (approved):

bitwalker:
Thanks!

PR #711, opened 2022-05-03
Expose TimezoneInfo.format_offset/1

nathanl:
### Summary of changes

Expose `TimezoneInfo.format_offset/1`

Previously, this was used only for inspection, but it can be useful for
user-facing elements such as time zone dropdowns.

### Checklist


bitwalker (approved):

bitwalker:
Thanks!

PR #714, opened 2022-06-20
🐎 Make Timex.Duration.parse ~2x faster

ulissesalmeida:
It's very important for us at Duffel speedup the duration
parsing. We're were investigating this function and we
noticed a big number of calls to `String.contains?` that is
a O(n) order algorithm in a loop.

So we did a quick experiment on putting a flag on the number
type while reading the chars. So it assumes that is a integer,
until find a dot, where it changes the type to float. Then, it
parses using Float or Integer depending on the flag. Avoiding
the String.contains calls.

The results were quite interesting:

```
Before
Timex.Duration.parse		500000	3322687 ~6.65µs/op

After
Timex.Duration.parse		500000	1710993 ~3.42µs/op
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added/Changed
 
+- Changed `Timex.Duration.Parse` to be 2x faster
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


paulogr:
fasdfasdf

Issue #718, opened 2022-07-25
Compile warning when using gettext 0.20.0


ypconstante:
### Steps to reproduce

- Add timex 3.7.8 and gettext 0.20.0 to project dependencies
- run `mix deps.get && mix deps.compile`
- gettext will show a warning

### Description of issue

gettext deprecated the compiler on version 0.20.0, and [since timex adds gettext to compilers list](https://github.com/bitwalker/timex/blob/ed146eee9912e5f2fdbff71c308d3630fa67940a/mix.exs#L15), a warning will always be shown.

gettext compile change: https://github.com/elixir-gettext/gettext/commit/7dc9e4502e4323070c0ba1dbda504d0b829fb979?diff=split#diff-38ce6818891a446a1c2d89e7395759ce1c056424734593853cc8e0b714c39f32

pecampelo:
I have been having the same issue, in the same versions you have reported:

`timex 3.7.8`
`gettext 0.20.0`

 It actually does not end compilation, and just hangs forever.

bitwalker:
Fixed as of 3.7.11

PR #723, opened 2022-09-03
Remove compilation warning

conradfr:
### Summary of changes

Remove the compilation warning.

-----

warning: the :gettext compiler is no longer required in your mix.exs.

Please find the following line in your mix.exs and remove the :gettext entry:

    compilers: [..., :gettext, ...] ++ Mix.compilers(),

-----

### Checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -12,7 +12,7 @@ defmodule Timex.Mixfile do
       package: package(),
       deps: deps(),
       docs: docs(),
-      compilers: [:gettext] ++ Mix.compilers(),
+      compilers: Mix.compilers(),
       test_coverage: [tool: ExCoveralls],
       elixirc_paths: elixirc_paths(Mix.env()),
       preferred_cli_env: [
--- a/mix.exs
+++ b/mix.exs
@@ -7,7 +7,7 @@ defmodule Timex.Mixfile do
     [
       app: :timex,
       version: @version,
-      elixir: "~> 1.8",
+      elixir: "~> 1.11",
       description: description(),
       package: package(),
       deps: deps(),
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -12,7 +12,7 @@ defmodule Timex.Mixfile do
       package: package(),
       deps: deps(),
       docs: docs(),
-      compilers: [:gettext] ++ Mix.compilers(),
+      compilers: Mix.compilers(),
```
florish, line 15: I think this line could be removed, as `Mix.compilers()` seems to be the default value if nothing is specified, given that I understand [the documentation](https://hexdocs.pm/mix/Mix.html) correctly.

florish:
This fixes https://github.com/bitwalker/timex/issues/718

florish:
To be sure: I'm not involved in maintaining `timex`, I also get this error every time I run `mix compile` and was happy to find both an issue and a PR addressing this 👍 . Just added a single comment and linked this PR to an issue. Thanks!

mix.exs
prihandi: I think better to check Elixir version first Mix.compilers(), since removal of mix task compiler only apply to version 1.11+.

```suggestion
      compilers: compilers_opts(),
      
   defp compilers_opts() do
     case Version.compare(System.version(), "1.11.0") do
        :lt -> [:gettext] ++ Mix.compilers()
        _ -> Mix.compilers()
      end
   end
```

florish: @prihandi Thanks for your comment. Removing the `:gettext` compiler is a fix for _Gettext_ deprecating its usage, not _Elixir_ issuing a deprecation warning. Your suggested code change therefore should not be applied as it does not fix any backward compatibility issues.

In case I'm missing something, let me know!

conradfr:
The warning comes from gettext v0.20.0, which itself has elixir 1.11 has a minimum version.

Timex is currently at elixir: "~> 1.8" so it will be up to the maintainer to decide if he wants to bump versions.

bitwalker (approved):

Issue #731, opened 2023-01-17
Timex.shift/2 with time-shift always converts precision to :microsecond in Elixir v1.14+


enerick:
### Steps to reproduce

with Elixir v1.13
```elixir
iex(1)> now = ~U[2023-01-01 10:20:30Z]
~U[2023-01-01 10:20:30Z]
iex(2)> Timex.shift(now, hours: 1)
~U[2023-01-01 11:20:30Z] # OK
```

with Elixir v1.14
```elixir
iex(1)> now = ~U[2023-01-01 10:20:30Z]
~U[2023-01-01 10:20:30Z]
iex(2)> Timex.shift(now, hours: 1)
~U[2023-01-01 11:20:30.000000Z] # precision changes unnecessarily
```

### Description of issue

- Elixir v1.14 changes DateTime.add/4 behaviour to preserve the maximum precision of given parameters.
   - commit: https://github.com/elixir-lang/elixir/commit/5a583c753b96865a7cdec2fb4c1ab9c96b836d24
   - relates to: https://github.com/elixir-lang/elixir/issues/12303
- Timex.shift/2 uses DateTime.add/4 with `:microsecond` precision to apply time-shift result.
   - https://github.com/bitwalker/timex/blob/e1c4d424a5c7840f63efd33cd70e3b5d4f9bfaa1/lib/datetime/datetime.ex#L450

This causes unnecessary precision conversions in Elixir v1.14 and later.

aniravi24:
For me this has been happening specifically in 1.14.3, not any of the other 1.14 versions.

aniravi24:
It looks like Elixir adds the precision to datetime and naive_datetime unlike before where it didn't merge it
https://github.com/elixir-lang/elixir/compare/v1.14.2...v1.14.3#diff-b83c142c2405080dc98ca11f1da9784fe7199a376c974e456a70b86c14bfd1a9R1547

aniravi24:
It might be this commit
https://github.com/elixir-lang/elixir/commit/5a583c753b96865a7cdec2fb4c1ab9c96b836d24

railsmechanic:
Same happend to NaiveDatetime, as it uses 

```elixir

PR #741, opened 2023-04-13
Fix: preserve datetime precision after Timex.shift/2

ucwaldo:
### Summary of changes

Resolves https://github.com/bitwalker/timex/issues/731

Unfortunately this broke with https://github.com/elixir-lang/elixir/commit/5a583c753b96865a7cdec2fb4c1ab9c96b836d24 which was released with Elixir 1.14.3

As `Timex.shift/2` calculates the total microseconds of the shift options and then uses `DateTime.add/4` to apply, Elixir returns a new datetime with the highest precision it finds (microseconds) opposed to keeping the original precision.

Elixir 1.13.4:

```
iex(3)> DateTime.utc_now() |> IO.inspect() |> DateTime.truncate(:second) |> IO.inspect() |> Timex.shift(minutes: 1) |> IO.inspect()
~U[2023-04-13 09:56:10.136274Z]
~U[2023-04-13 09:56:10Z]
~U[2023-04-13 09:57:10Z]
```

Elixir 1.14.4:

```
iex(1)> DateTime.utc_now() |> IO.inspect() |> DateTime.truncate(:second) |> IO.inspect() |> Timex.shift(minutes: 1) |> IO.inspect()
~U[2023-04-13 09:55:16.405357Z]
~U[2023-04-13 09:55:16Z]
~U[2023-04-13 09:56:16.000000Z]
```


ucwaldo:
@bitwalker is this a change you'd be considering for Timex?

lib/datetime/datetime.ex
```diff
@@ -480,6 +484,13 @@ defimpl Timex.Protocol, for: DateTime do
       err
   end
 
+  defp retain_precision(
+         %DateTime{microsecond: {ms, _prevision}} = new_datetime,
```
azizk, line 488: Just noticed a minor typo: `_prevision` should be `_precision` :slightly_smiling_face:

lib/datetime/datetime.ex
```diff
@@ -447,13 +447,17 @@ defimpl Timex.Protocol, for: DateTime do
...
+          |> retain_precision(datetime)
 
         %DateTime{} = datetime ->
           datetime
 
-        {{ty, _, _}, %DateTime{} = orig} when ty in [:gap, :ambiguous] and shift != 0 ->
-          DateTime.add(orig, shift, :microsecond, Timex.Timezone.Database)
+        {{ty, _, _}, %DateTime{} = original} when ty in [:gap, :ambiguous] and shift != 0 ->
+          original
+          |> DateTime.add(shift, :microsecond, Timex.Timezone.Database)
```
azizk, line 459: Maybe it's a good idea to create a function for adding microseconds while keeping the precision? Something like:

```elixir
original |> add_with_precision_preserved(shift)
```

lib/datetime/datetime.ex
```diff
@@ -447,13 +447,17 @@ defimpl Timex.Protocol, for: DateTime do
...
+          |> retain_precision(datetime)
 
         {{ty, _a, _b} = amb, _} when ty in [:gap, :ambiguous] ->
           amb
```
azizk, line 463: I would get rid of the case statement below as it seems completely unnecessary, but that may be outside of the scope of this PR. :shrug: :sweat_smile:

ftes:
@bitwalker Do you have any time plans for reviewing/merging this PR?

"I don't have time" is also a totally fine answer that helps us plan accordingly. Thank you for your efforts!

petermueller:
Thank you doing this work. I've been unable to complete an upgrade to 1.14 as a huge number of our legacy database tables were created w/ resolution only down to the second.
We thankfully have a shim layer where we can perform this for as well, but would love to see this merged to make our testing a little easier

edennis:
Any chance of getting this merged? 🙏🏻

PR #763, opened 2024-09-22
simplify months attribute value generation

Eiji7:
No need to map after indexing. Setting offset or function in second argument of [Enum.with_index/2](https://hexdocs.pm/elixir/Enum.html#with_index/2) is shorter and faster.

### DIFF
--- a/lib/timex/constants.ex
+++ b/lib/timex/constants.ex
@@ -39,7 +39,7 @@ defmodule Timex.Constants do
         "November",
         "December"
       ]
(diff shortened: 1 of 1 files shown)


Eiji7:
The `Enum.with_index/2` function (with an integer as a second argument) was added to `Elixir` in version `1.2.0-rc.0`, see:
https://github.com/elixir-lang/elixir/blob/v1.2.0-rc.0/CHANGELOG.md#v120-rc0-2015-12-09

Is there any reason to not merge this PR?

bitwalker (approved):

PR #766, opened 2024-10-21
Fix tzdata terminating host application v1.1

shegx01:
### Summary of changes
Fix issue with tzdata v1.1.2 terminating the host application server

### Checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -55,7 +55,7 @@ defmodule Timex.Mixfile do
 
   def deps do
     [
-      {:tzdata, "~> 1.1"},
+      {:tzdata, "~> 1.1.2"},
       {:combine, "~> 0.10"},
       {:gettext, "~> 0.20"},
       {:ex_doc, "~> 0.13", only: [:docs]},


corroded:
@shegx01  could it be because of this? https://github.com/lau/tzdata/issues/148

I keep encountering this in our logs - but it shouldn't terminate your host application server unless it fails to report errors or something? 🤔

kaylenmistry:
> @shegx01 could it be because of this? [lau/tzdata#148](https://github.com/lau/tzdata/issues/148)
> 
> I keep encountering this in our logs - but it shouldn't terminate your host application server unless it fails to report errors or something? 🤔

Does not terminate the application, however would be nice to suppress this error log. Any chance we can proceed with this change? Thanks!

bitwalker (approved):

bitwalker:
It isn't clear to me why this is actually a proper fix - you should be able to just `mix deps.update tzdata` to get `1.1.2+` with the current dependency constraint. Have you tried that?

shegx01:
This is how we resolve the issue but for those that unlock dependencies or remove the deps.lock and reinstall, this issue would crept back in.

bitwalker:
Fixed in latest release

